PR #1, opened 2018-01-12
Update README.md

sblunt:
Added description to readme


semaphoreP:
Less exclamation marks would be good. If you update that, then it should be ok.

PR #13, opened 2018-01-20
Added Keplerian orbital solver

semaphoreP:
I added the kepler.py module and a test for it. I'm using Rob De Rosa's eccentricity solver, which seems to produce different answers than James' of around ~10 microarcseconds. Not sure why, but something to look into in the future.

### DIFF
--- a/orbitize/kepler.py
+++ b/orbitize/kepler.py
@@ -25,9 +25,9 @@ def calc_orbit(epochs, sma, ecc, tau, argp, lan, inc, plx, mtot, mass=0):
         mass (float): mass of this body [Solar masses]. For planets mass ~ 0
 
     Return:
-        raoff (np.array): array of RA offsets from COM for each epoch
-        deoff (np.array): array of Dec offsets from COM for each epoch
-        vz (np.array): array of radial velocities of the body relative to COM for each epoch
+        raoff (np.array): array of RA offsets between the bodies (origin is at the other body)
+        deoff (np.array): array of Dec offsets between the bodies
+        vz (np.array): array of radial velocity offset between the bodies
     """
 
     ndates = np.size(epochs)
--- a/orbitize/kepler.py
+++ b/orbitize/kepler.py
@@ -1,7 +1,7 @@
 """
 This module solves for the orbit of the planet given Keplerian parameters
 """
-import numpy as np 
+import numpy as np
 import astropy.units as u
 import astropy.constants as consts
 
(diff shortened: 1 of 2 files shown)


sblunt:
Thanks for getting `kepler.py` in shape, @semaphoreP! The code looks great. Two small comments:

1. Hm, strange that it produces different answers. Have you asked Rob why that might be? I can volunteer to take the lead on that conversation if you haven't.

2. Also, `calc_orbit()` should be able to take in arrays of `sma`, `ecc`, etc for OFTI to work. I can also volunteer to make sure it works for those inputs, unless you'd like to take care of it (it would be good for me to get to know kepler.py a little better).

henry-ngo:
It looks like De Rosa's solver as implemented is an analytical solution, right? It's been awhile but I am pretty sure that Kepler's equation cannot be exactly solved analytically, so any analytical solution is only an approximation. Maybe this is the source of the difference?

In De Rosa's IDL code (as linked in the docstring), after the analytical solution is solved, there is a comment block that says

>  ;; Now get more precise solution using Newton Raphson method
>   ;; for those times when the Kepler equation is not yet solved
>   ;; to better than 1e-10
>   ;; (modification J. Wilms)

1e-10 radians is 20 microarcseconds. Maybe that's just a coincidence, though. I remember from my dynamics class that one of the things we had to do was implement a Newton solver for the Kepler Equation. 

This is going to be much slower though but whether it's too slow for the many many times orbitize is going to solve the equation is not intuitive to me (I think yes, but have no data; however, if James' solution is also numerical then maybe it's okay?). 

Suggestion: can you confirm that using a numerical solution with tolerances down to 1 microarcsecond in the eccentric anomaly will match James' solution? If so, then we'll at least know the source of the discrepancy and we can probably default to analytical solution with a keyword to specify tolerance for users who don't care about run-time and really care about accuracy?

semaphoreP:
There's both an analytical and iterative solver. I'm mostly right now comparing the iterative solver from James and Rob's codes. Both have a tolerance value for how good the iterative solution needs to be, but both seem to converge to different answers at the 10 microarcsecond level (as I increase the tolerance to smaller and smaller values). We can discuss more on Slack if you'd like.

henry-ngo:
Slack discussion summary thus far: I didn't see that there's a branch in `_calc_ecc_anom` that only uses the analytical solver when e>0.95. New unit tests does not seem like there's a problem with the eccentric anomaly calculation, other than the iterative solver yielding 1e-8 precision when tolerance set to 1e-9.

semaphoreP:
I am going to merge this feature now. We can add @sblunt's TODO on the supporting multiple sets of orbital parameters in a future pull request. I don't want this pull request to go left hanging for too long.

PR #14, opened 2018-01-22
Adding Travis CI

sblunt:
- created travis.yml file that tells Travis CI how to test `orbitize`
- updated `contributor_guidelines.md` with unit testing instructions
- added cool badges to `README.md`


sblunt:
@semaphoreP @henry-ngo this branch is ready for integration pending your approval. 
- badges are all correct now.
- after a lengthy battle with the `python-cov` battle, I came out victorious! This branch now runs `pytest` tests and sends the results to coveralls.io. **Pro Tip: ** there's a bug with `python-cov` that requires the `orbitize/tests` subdirectory to have an `__init__.py` file for it to run properly.

semaphoreP:
Woot! I had the same problem fighting against the build system before I got the tests working. If you want to investigate it further, I know for pyklip we got it all working without `pytest-cov`. I used a pytest.ini and .converagerc file to configure everything. That was also with a different build system, and not travis although I don't think they should be that different. Anyways, it works now and making the tests directory a subpackage is not a big deal.

PR #18, opened 2018-01-31
Improve read input

### LINKED ISSUE #5: Implement read in data from text files
At least support CSV for now. Include some documentation on what the format should be.

henry-ngo:
Extended the read_input.py module to be able to read and write input tables in the format that orbitize expects (i.e. quant1/quant2/type) instead of astrometric measurements (ra/dec/sep/pa etc.). This allows a user to give input in more intuitive astrometric measurements as well as the specific input type that orbitize expects. The write functions allows the conversion from measurements to orbitize format to be saved and reloaded.

I also added tests for these functions that now actually check if the expected values are read in from the files in test_read_input.py (previous "test" just prints out the file). 

I would appreciate it if someone could check (thanks in advance!):
1. Whether I am using the docstrings correctly for the read_input.py functions. I tried to follow Jason's advice to make a table but I'm not sure if this is the way it should look. Maybe it should be an "example" instead?
2. I'm new to pytest so let me know if I did something silly!

Finally, I think we should look at updating and modifying these input functions once the samplers are implemented and we know what information/types will work best. But once this PR is merged, I'll close #5 as basic functionality is now here and we can open new issues for any further upgrades.


semaphoreP:
Yes, I agree that once we get some basic stuff running and try using this interface, we'll have better ideas for what to change. 

Here's what the docstring looks like in the doc page (I just ran `make html` in the docs folder). 
image

I would say perhaps using a code block example with ascii text might be more straightforward. Currently we have nicely formatted table columns, but the user probably is more interested in how to format their ASCII file correctly. 

Otherwise, everything else you did/mentioned in this PR seems perfectly reasonable to me.

henry-ngo:
Okay, updated the docstring to be more readable. Not sure what one of my lines has a gray background. Although I guess highlighting that line in the docs isn't a bad idea. Here's an image to show what I mean: 

image

sblunt:
Looks awesome, Henry. Sorry it took me so long to review!

Issue #16, opened 2018-01-31
Pass in multiple sets of orbital parameters into kepler.py


semaphoreP:
Especially for OFTI, it'd be great to pass in arrays of orbital parameters into `kepler.calc_orbit`. 

Basically the `epochs` argument can be of length M and the `sma`, `ecc`, etc arguments can be of length N, and the return values (e.g., `raoff`) should be of dimension NxM.

This will requiring formatting the code to angle scalars/1-D vectors at the same time, as well as 1-D/2-D vectors at the same time. It would be great to avoid code duplication. I _think_ this is possible.

henry-ngo:
Okay so the branch vectorize_kepler has some basic functionality for passing in arrays of orbital parameters now! From my commit notes:

For calculations with 2-D arrays, currently looping through the "epochs" dimension (number of dates) because 1) it's likely to be smaller and 2) it prevents a lot of issues where sometimes we need to treat a keplerian element as a vector vs. a scalar.

The exception is the calculation of the eccentric anomaly, which is still done with a single value of eccentricity. Further upgrade changing this calculation can speed up the code.

For now, assume that the "mass" keyword still only points to one single value. Would not be hard to extend this to an array but not sure if that's what would be used? Let me know!

At the end, if np.squeeze() is used to reduce extra dimensions.

A few clarification questions.

1. Does the code need to take an array for "mass" (the mass of the orbiting body) or would it just be a single value?

2. Is this the "vectorization" you had in mind? If the single for-loop for 2-D calculations are too computationaly costly, I can probably map everything onto a 1-D array and then reshape to 2D again? That would make the code less readable, though. But it could be faster that way.

sblunt:
Thanks again for working on this, Henry. To answer your questions:

1. Yes, we will eventually want to take in a value and some uncertainties on the system mass, so it would be good to extend that to an array. Will the code as-is just work with mass as an array (in which case we'd just update the docs), or are there some modifications to make? 
2. Yes, this is definitely the vectorization I had in mind. The computational speedup will be awesome for OFTI.

sblunt:
One more small thing: in the docstring for `calc_orbit`, there's an incomplete line:

`mtot (np.array): total mass [Solar masses]. Note that this is`

henry-ngo:
Just a tiny modification needed for an array input for mtot, plus finishing off that doc line. I'll update it soon!

henry-ngo:
[...]

PR #19, opened 2018-02-02
Vectorize kepler

henry-ngo:
This upgrade addresses issue #16 and allows kepler.py's calc_orbit to handle arrays as well as scalars as input for orbital elements so that we solve many orbits simultaneously.

Note that calc_orbit takes in epochs which has size n_dates and orbital elements which has size n_orbs. For most cases, n_dates << n_orbs so almost all of the loops in calc_orbit is a loop through the epochs. The exception is the eccentric anomaly solver, which can only solve for one eccentricity at a time, so it is a loop through each set of orbital elements. Something to keep in mind if we need further speedups.

So far, a speed test with 100,000 simultaneous orbits on this branch takes 3.7 seconds. Running the old version (scalars for orbital elements) 100,000 times takes about 90 seconds.

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -1,8 +1,5 @@
 Sarah Blunt -- collaboratively designed basic API
 
-Jason Wang -- collaboratively designed basic API
+Jason Wang -- collaboratively designed basic API, wrote kepler.py
 
-Henry Ngo -- wrote code to read inputs
-
-
-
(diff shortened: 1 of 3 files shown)


sblunt:
Looks good, Henry. Thanks again for doing this (and +1 for profiling the code, too). Like Jason said, it might be good to eliminate some of the `for` loops in the future, but that's a job for another day. It's awesome that the solver is so much faster now!

PR #21, opened 2018-02-12
Eanom optimizations

rjderosa:
A few modifications:

- Element, parallax and mtot vectors are now converted into (n_orbs, n_dates) arrays. A bit clunky. The alternative is to use [:, None] indexing, although that would have to be put everywhere in the code, so maybe this is neater?
- manom now wrapped between 0->2pi. This significantly changes the output for high eccentricity orbits.
- raoff, deoff and vz calculation vectorized
- eanom calculation vectorized
- Split the newton solver into its own function.
- Various modifications for consistency, and removed some unused variables.

### DIFF
--- a/orbitize/kepler.py
+++ b/orbitize/kepler.py
@@ -110,20 +110,30 @@ def _calc_ecc_anom(manom, ecc, tolerance=1e-9, max_iter=100):
     Code from Rob De Rosa's orbit solver (e < 0.95 use Newton, e >= 0.95 use Mikkola)
 
     Args:
-        manom (np.array): array of mean anomalies, shape (n_orbs) or (n_orbs, n_dates)
-        ecc (float/np.array): eccentricity, either a scalar or shape (n_orbs, n_dates)
+        manom (float/np.array): mean anomaly, either a scalar or np.array or any shape
+        ecc (float/np.array): eccentricity, either a scalar or np.array of the same shape as manom
         tolerance (float, optional): absolute tolerance of iterative computation. Defaults to 1e-9.
         max_iter (int, optional): maximum number of iterations before switching. Defaults to 100.
     Return:
-        eanom (np.array): array of eccentric anomalies
+        eanom (float/np.array): eccentric anomalies, same shape as manom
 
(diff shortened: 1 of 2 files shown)


sblunt:
@rjderosa do you want to merge this PR into `master` now or iterate for a while longer?

henry-ngo:
I'm going to add some commits to fix that missing docstring text. 

Also, to allow for an array input for `mass`, as suggested above.

rjderosa:
I think i've finished tinkering with the eccentric anomaly stuff. It should work for when `ecc` and `Manom` are both scalars, when they are both vectors, and when `ecc` is a scalar and `Manom` is a vector. I don't think you'll ever be in a situation where you have a single mean anomaly with multiple eccentricities?

It can probably be tidied up a bit, especially if you wanted to do something more fancy that the transposing/tiling at the top to make all the arrays into shape (n_orb, n_epoch).

henry-ngo:
Great. I'm about to commit the changes I mentioned above to this branch too.

rjderosa:
Could we not compute ka and kb in the same way for both mass==0 and mass!=0? When it's massless `ka` will just be zero.

henry-ngo:
@rjderosa Good point. let me check. I just kept the same structure as before since I haven't had much experience calculating RV semi-amplitudes and thought it was separate for a reason.

henry-ngo:
Yes, you're right---passes all the tests! Updated now!

rjderosa:
Looks as though the second equation is all you need for `ka`. We could add the functionality to fit SB2s, `k2` would just be the same equation but with the primary mass rather than the secondary (and a negative sign).

rjderosa:
We should probably rename `m2` to something that doesn't sound like it's the mass of the secondary! Perhaps `m1`, or `ma`?

henry-ngo:
I called it m1 now and put a comment to say it's the mass of the primary star. Given the purpose of this package, I think we can assume that in most cases, m1 will be a star (but I suppose we can use this code to calculate a binary(!) planet haha)

rjderosa:
Or black hole binaries!

PR #24, opened 2018-04-09
Cpp kepler module

devincody:
Added multithreaded C++ solver for the Kepler module. Defaults to Numpy Kepler if C++ code not built. Used Cython to wrap the C++ functions. The number of threads is a function option.


sblunt:
@devincody this code now builds properly on my computer. Let me know once you've tested it on a Windows machine, and then we'll be good to pull. 

Also, note for posterity that multithreaded option was removed due to difficulty installing. We'll parallelize `orbitize` at the sampling stage.

PR #32, opened 2018-06-14
pytest requires >=3.0.0

devincody:
the pytest.approx() function was added in 3.0.0


sblunt:
Sounds good. Merging.

Issue #34, opened 2018-06-29
Add label for each parameter in the systems class


semaphoreP:
This should be a field in the `System` class that lists the labels (e.g., 'a', 'e') for each parameter so we (humans) know which parameter each is. We could consider using 'a1', 'e2' to help separate multiple planets.

henry-ngo:
Noticed that `System` currently has this description of parameter ordering for priors in its docstring

```
    Priors are initialized as a list of orbitize.priors.Prior objects,
    in the following order:

        semimajor axis b, eccentricity b, AOP b, PAN b, inclination b, EPP b, 
        [semimajor axis c, eccentricity c, etc.]
        mass, parallax

    where `b` corresponds to the first orbiting object, `c` corresponds
    to the second, etc. 
```

We should have a uniform way of numbering objects/planets, whether it's letters or numbers.  I would suggest using numbers instead of letters now to denote objects and maybe even an underscore to avoid confusion with exponents (i.e. using 'e_2' instead of 'e2' makes it more clear that it is the eccentricity of the second object, rather than the eccentricity squared)

PR #35, opened 2018-07-03
Improve system+results

sblunt:
- added list of labels to System class (issue #34) and added `results` attribute to System class
- added unit test for System labels
- fleshed out docstring of Results class

### DIFF
--- a/orbitize/__init__.py
+++ b/orbitize/__init__.py
@@ -1,4 +1,5 @@
 import sys
+import types
 
 __version__ = '0.1.0'
 
--- a/orbitize/results.py
+++ b/orbitize/results.py
@@ -1,9 +1,31 @@
 class Results(object):
     """
-    Stores accepted orbital configurations from the sampler
+    A class to store accepted orbital configurations from the sampler
+
+    Args:
+        post (np.array of float): MxN array of orbital parameters
+            (posterior output from orbit-fitting process), where M
+            is the number of varying orbital parameters in the fit, 
+            and N is the number of orbits generated [None].
+        mass_err (float [optional]): uncertainty on ``system_mass``, in M_sol
+        plx_err (float [optional]): uncertainty on ``plx``, in arcsec
+
+
+    The `post` array is in the following order:
+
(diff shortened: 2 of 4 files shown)


semaphoreP:
Looks good! Just one quick thing to change: in system.py System.__init__(results=None) not results=[], as you should basically never have the default value of function arguments as objects (see https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument); generally only numbers, strings, and booleans are ok for default values as they are stored as their values in the function and not a pointer.

@henry-ngo, we should also consider adding a function in the MCMC sampler to inspect the MCMC chains (either visually or by some metrics) and allow the user to chop off additional steps as "burn-in" if part of the MCMC chain hasn't converged yet. And do this before we create the results object? Or maybe update the results object with the updated chain.

henry-ngo:
I think this looks good. Just one comment:

To make the docstrings more readable, we should write what AOP, PAN and EPP is short for, probably just after the text that says "1 correspond to the first object" etc. I am sure the tutorial will have more to say but AOP/PAN/EPP are hard to Google while the full names can be Google-able by someone new to orbit fitting. 

Also, these angles are often known as their greek letters or other names, so writing it out will be more clear to every audience.

As an example: PAN is something to do with the ascending node right? I had always thought the angle was called the "longitude of ascending node", which should be LAN, not PAN. Is PAN a different name for this angle (what's the P?) or is it a typo?

@semaphoreP : I agree with you re: the addition to the MCMC Sampler. My first thought was that we should save everything to the Results object, and then Results might have a method that allows the user to inspect and chop off the burn-in. But then I remembered that Results is a generic object, not just for MCMC results.

So, I think all the functions for inspecting/chopping-off/adjusting burn-in should live in the PTMCMCSampler object. Changes to the chain due to these adjustments should then update the Results object, as you suggest. I think we should be able to do this both before and after the Results object is created. That is, the inspect method should load up the Results object and then the chop methods can update the Results object as necessary.

In any case, those changes belong in a different PR, I think.

@sblunt : I think you can go ahead and merge this after the two minor comments from Jason and myself are implemented :)

sblunt:
Thanks for the great comments, all! Changes have been implemented.

Issue #30, opened 2018-04-26
Make sure everything is pickleable and implement multiprocessing for MCMC


sblunt:

henry-ngo:
Great! :)

semaphoreP:
Ok, I think tests/test_mcmc.py has the bare-bones code needed to get the MCMC sampler running. Let me know if you have any questions/issues getting it running. If system.py is pickleable, hopefully, it should run parallelized by specifying number of threads when creating the emcee sampler!

sblunt:
Addressed in PR #39.

PR #39, opened 2018-07-17
Multi-threading support for MCMC samplers

henry-ngo:
sampler object's initialization can specify `nthreads` for parallelization. 

Since `emcee` will remove their PT sampler in the next release, I've added the Ensemble Sampler from `emcee` as an alternative (a good replacement might be `ptemcee`)

Addresses Issue #30


henry-ngo:
Okay, the new version of emcee actually deletes their PTSampler and instead checks if you have ptemcee installed and if so, imports that instead.

sblunt:
Thanks, @henry-ngo! Milestone!

PR #42, opened 2018-07-20
Switch ptemcee

henry-ngo:
Minor changes to fully embrace ptemcee for PT sampling


henry-ngo:
okay as per slack chat, we'll leave the PR at this commit and not add to the `install:` block in .travis.yml

PR #46, opened 2018-08-15
Kepler module implemented in C

devincody:
Implementation of kepler module in C. I've installed it successfully on my Windows Subsystem for Linux (Ubuntu) and Windows terminals. Includes profiling tool and is compatible with whichever archaic C compiler that Travis uses (pre C99).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,9 @@ __pycache__/
 # C extensions
 *.so
 
+# Cython generated files
+_kepler.c
+
 # Distribution / packaging
 .Python
 env/
@@ -45,6 +48,7 @@ nosetests.xml
 coverage.xml
 *.cover
 .hypothesis/
+Profile.prof
 
 # Translations
 *.mo
@@ -99,3 +103,6 @@ ENV/
 
 # mypy
 .mypy_cache/
+
+# utility scripts
+build.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ before_install:
     - sudo ln -s /run/shm /dev/shm
 
 install:
-    - conda install --yes python=$TRAVIS_PYTHON_VERSION pip numpy scipy
+    - conda install --yes python=$TRAVIS_PYTHON_VERSION pip numpy scipy cython
     - pip install pytest pytest-cov coverage coveralls
     - pip install . --upgrade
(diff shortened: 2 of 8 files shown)


henry-ngo:
I think this generally looks good. My C is a little rusty, but seems fine. My only comments are:

1. the C implementation of the solver returns eanom = -1 when it fails to solve, but the comment says that the numpy version doesn't do this (which I checked and it appears to just return whatever value it was at when it ran out of iterations). I think we should make the two parts behave the same, i.e. change the numpy version to do the same. Can you add that change to the numpy version too, please, @devincody ?

2. Make sure you add your contribution to contributors.txt :)

rjderosa:
Returning a float sounds like a bad idea, especially if E is wrapped (e.g., -1 would become 2pi-1).

Should we have it raise an error if it fails to converge? I've yet to find a combination of (M, e) that it fails on, but you never know.

henry-ngo:
My concern with raising an error would be in cases where you don't want the whole thing to come to a halt if one call to kepler.py fails, right? For example, I would imagine in most cases, you might want to flag it so that other code can decide how to treat this case (e.g. run it through a different solver? rerun with more iteration? etc.)

Perhaps we should use `None` as a flag? Or `np.nan`?

sblunt:
That's a good point. I think np.nan is good. That behavior is consistent with what we want to happen if a parameter wanders outside of the allowed prior space.

rjderosa:
> the C implementation of the solver returns eanom = -1 when it fails to solve, but the comment says that the numpy version doesn't do this (which I checked and it appears to just return whatever value it was at when it ran out of iterations)

Oh, and I should correct this. If the python version was ever to run out of iterations, it calls the analytical solver, rather than simply returning the value at the last iteration.

henry-ngo:
Ah okay, thanks for the correction @rjderosa

devincody:
Just saw these comments, but it seems like you all have figured it all out without me. The python _newton_solver calls the _mikkola_solver when it finds an orbit that doesn't converge. The _c_newton_solver does the same but cannot call the _mikkola_solver directly so it flags problematic orbits with -1 which are then lumped in with the high eccentricities and solved.

sblunt:
Looks great, @devincody. Thanks for all your work on this!

Issue #45, opened 2018-08-14
Ordering of orbital elements


semaphoreP:
We should make `system.py` and `sampler.py` use the same ordering of orbital elements, and make it less arbitrary. Henry suggested the following order that makes more sense:

> a (semi-major axis)
> e (eccentricity)
> i (inclination)
> omega or w (arg of pericentre)
> Omega or O (long. of ascending node)
> tau (or Mean anomaly or whatever equivalent coordinate one uses to denote a specific time from a specific point in orbit)

Code for both samplers as well as system.py need to be modified to support this.

sblunt:
How are we doing on this @henry-ngo? You addressed a few of these in your recent PR, right?

sblunt:
Thanks @henry-ngo !!

sblunt:
Ok yeah I'm running into this issue trying to merge Isabel's OFTI branch into master. Let me know when you've fixed this in the places you mentioned, then I can make sure it's fixed in `sampler.OFTI`. We also need to make sure it's fixed in the C implementation of `calc_orbit()` @devincody.

henry-ngo:
Oh sorry for not being clear, PR #55 fixes it in all of these places! 

I also checked the kepler.c and how the C implementation appears in kepler.py and if I understood the code correctly, the C newton solver is called on line 151 and it doesn't use the list of orbital elements. Instead, you need to pass the arrays for `manom` and `ecc` only so we should be fine there!

sblunt:
Excellent! My bad, I didn't actually check-- I just remembered that's another place we'd have to remember to fix this.
Ok, so I'll go ahead and fix the OFTI ordering and hopefully that will work!!

sblunt:
Addressed in PR #55 and PR #48.

PR #55, opened 2018-09-05
Fix orbel order

henry-ngo:
This should correct the order of orbital elements in all places, as discussed in #45 

In most places, this amounts to switching the positions of `inc` and `epp`


sblunt:
Cool, looks great. Thanks again for fixing this, @henry-ngo .

Issue #20, opened 2018-02-12
Speed up `numpy` implementation of `kepler.py`


sblunt:
- turn `for` loops into vector array operations

henry-ngo:
I was just reviewing this too. I think we have already addressed the main point in the issue awhile ago (turn for loops in vector ops). But last week Rob mentioned another big speedup if we pass `manom` as an array. I think that is good but it does break some parts of the code, which I'm discussing with him on Slack now in a thread from his comment.

So I think we can keep this issue open until we resolve that.

sblunt:
Ok, sounds good.

sblunt:
Addressed in PR #58

PR #58, opened 2018-09-06
Force iterative part of numpy kepler solver to work with np.array 

henry-ngo:
This is done in case user passes astropy.Table Column objects (which would be a lot slower)

This should wrap up #20 if @rjderosa agrees that it addresses the concern!


sblunt:
Looks good to me! Let's wait to see if @rjderosa is happy, then we can pull in this fix. Thanks @henry-ngo!

PR #65, opened 2018-10-02
Save results

henry-ngo:
Implemented methods in results.py to save/load results objects as HDF5 and FITS files.


henry-ngo:
This isn't a current v1.0 issue, however, I should have implemented these methods along with the other "results.py" v1.0 milestones.

Issue #56, opened 2018-09-05
Fix implementation of stellar mass and parallax


sblunt:
Instead of defining separate variables for mass and parallax when they are fixed, let's just add them to `system.sys_priors` as floats instead of `Prior` objects

sblunt:
I should have documented our plan for this immediately after our telecon, but I forgot! Oh well. I think the plan is (correct me if I'm wrong):
- add a helper method in `system.py` that takes in the full list of parameters (`float`s & `Prior` objects) returns a list of just `Prior` objects. 

Are you good to finish that up @semaphoreP?

semaphoreP:
I don't know whether it's best to handle this in the system.py class or in the sampler.py class. But I can think about it, and go with which I think is better.

sblunt:
Ok, sounds good! Right, I forgot that we had discussed where to put this helper function.

semaphoreP:
Ok seems like there's two ways to go:

1. sampler.compute_model(params_arr) accepts either a full params_arr or a params_arr with len == number of non-fixed parameters, and a helper function to return a list of priors to sampler. This has the benefit that a user can use sampler.compute_model() passing in only the variables they are fitting for, and don't have to worry about also passing in the variables that are fixed. The only con is that this will complexify system.py and make MCMC/OFTI inconsistent: MCMC results will only return sampled parameters whereas OFTI will return all parameters (unless we also change OFTI).

2. Handle fixed parameters in an sampler.PTMCMC which has the benefit that the changes will be localized and have to handle less general cases. The major con is that for MCMC users, the MCMC chains will be missing the fixed variables, so we need to do something (e.g., add them into the chains?) in order for the output to be easily passable back into system.compute_model() for doing things like plotting the orbits or analyzing them in other ways. 

Thoughts?

henry-ngo:
Would both options result in the MCMC output chain only containing sampled parameters, or did I miss something?

sblunt:
I think I like option two better, because this does seem like an MCMC-specific issue. We could have an MCMC base class that has two helper methods for adding and removing fixed variables from arrays.

henry-ngo:
[...]

PR #69, opened 2018-10-18
Fix mass and parallax 2

semaphoreP:
Ok, I think the second try starting fresh was a lot more successful. 

- Re-implemented @sblunt's changes to system.py. 
- The results posteriors are always the same size, no matter if mass/plx are fixed. 
- Similarly, OFTI's samples are also fixed size. I think the slight increase in memory usage is negligible here. 
- MCMC now samples only the non-fixed parameters, but will return posteriors that include the fixed parameters. 
- Under this philosophy, it is up to the samplers to handle the fixed parameters gracefully (system.compute_model() requires mass and parallax as params as a result right now, since the system right now does not care about fixed parameters).
- Also, I fixed a bug where the Ensemble Sampler wasn't using the priors (was only calling the likelihood function).

Addresses #56

### DIFF
--- a/orbitize/sampler.py
+++ b/orbitize/sampler.py
@@ -295,9 +295,9 @@ class MCMC(Sampler):
         for prior in self.priors:
             # draw them uniformly becase we don't know any better right now
             # TODO: be smarter in the future
-            random_init = prior.draw_samples(num_walkers*num_temps)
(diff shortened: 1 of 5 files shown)


sblunt:
Looks great, @semaphoreP. Thanks for working through all these changes.

PR #78, opened 2018-10-30
Added logo to README and Docs

semaphoreP:
Here is how the Docs looks with the logo on the main page and as the favicon in the tab bar

image

And here is it in the Github README (not 100% sold on including it here)

image


sblunt:
Looks beautiful!

Issue #80, opened 2018-11-11
`Driver` should accept both properly inputted .csv files and `astropy.table.Table` objects


sblunt:
Currently, `Driver` expects a csv. We should make it flexible enough to accept either a csv or an `astropy.table.Table`. Thanks @logan-pearce for the suggestion!

henry-ngo:
`read_input.py` provides two possible read modes: 

`read_formatted_file` reads a CSV file that must be formatted with certain columns, typically values used by observers (e.g. delta RA/DEC, sep/pa, RV) and then generalizes this to a generic `orbitize` style.

`read_orbitize_input` reads any file with the generic `orbitize` style

`Driver` currently only uses `read_formatted_file`. To improve this, we should also allow it to use `read_orbitize_input` which would allow reading in of other filetypes created by `orbitize` as well. 

Reading in an `astropy.table.Table` directly could work too. The original intent of `read_orbitize_input` was to allow the user to use `write_orbitize_input` (another method of `read_input.py`) to save an `astropy.table.Table` to a file and then re-read it later. But allowing a shortcut makes sense too

sblunt:
Addressed in PR #82. Check it out, @logan-pearce!

logan-pearce:
Nice!  I like this feature, not having to write out a new file.  Great work.

PR #82, opened 2018-11-29
Driver & read_input fix

henry-ngo:
This update deprecates two functions in read_input: read_formatted_file and read_orbitize_input and combines their functionality in a new function: read_input

Also includes minor typo fixes and updates all calls to the deprecated functions to use the new functions instead. This update adds the deprecation module to label and track deprecated functions. Users will be asked to use read_file instead.

Currently, the above deprecated functions will still exist and serve as wrappers to read_input until version 2.0. New unit tests developed to test this functionality. These new unit tests have a decorator that will cause the tests to fail if we forget to actually remove the function in version 2.0.

Docs are also updated.

Finally, this also addresses Issue #80 with updates to Driver.py to allow input as either a file or an astropy.table.Table object. The input argument that used to take `filename` now accepts either a filename/path or a reference to an astropy.table.Table object. Unit tests for this have been added.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ before_install:
 
 install:
     - conda install --yes python=$TRAVIS_PYTHON_VERSION pip numpy scipy cython matplotlib
-    - pip install pytest pytest-cov coverage coveralls corner
+    - pip install pytest pytest-cov coverage coveralls corner deprecation
     - pip install . --upgrade
     - python setup.py build_ext -i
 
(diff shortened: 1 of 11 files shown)


henry-ngo:
Quick question:

For the deprecation notes, I said the two functions will be deprecated in v1.0.2 since that will be the first release after these changes are merged.

In general, should PRs include version increment, or should that be something @sblunt does as a commit to master after deciding when enough merges have been made to be the next release. I think the second option makes more sense but I'm happy to increment the version as part of this PR if you prefer. Just letting you know!

sblunt:
I will increment the version! Thanks for checking. 

The deprecation warnings are awesome. I love that the tests will fail if we forget to remove in future versions! So nifty! Thanks for working on this @henry-ngo.

PR #85, opened 2018-12-28
Do not include negative values in the Gaussian prior

sblunt:
This was causing unphysical (negative) stellar masses to be returned. Fixed.

### DIFF
--- a/orbitize/priors.py
+++ b/orbitize/priors.py
@@ -94,10 +94,10 @@ class GaussianPrior(Prior):
         """
         lnprob = -0.5*np.log(2.*np.pi*self.sigma) - 0.5*((element_array - self.mu) / self.sigma)**2
 
-        bad_samples = np.where(samples <= 0)[0]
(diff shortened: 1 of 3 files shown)


sblunt:
Ok, just added that keyword. Lmk what you think now!

orbitize/priors.py
```diff
@@ -101,7 +101,7 @@ def compute_lnprob(self, element_array):
 
         if self.no_negatives:
 
-            bad_samples = np.where(samples <= 0)[0]
+            bad_samples = np.where(element_array <= 0)[0]
```
henry-ngo: I know that there are issues with checking for equality when you have floating point numbers, so this probably makes no difference, but shouldn't the check here be for `element_array < 0` instead of `element_array <= 0`, just so it reads more clearly? Unless we do want to exclude zero specifically?

sblunt: Sure, that makes sense. I'll push that change to this branch. Thanks!

PR #86, opened 2018-12-28
Expand orbit plots

sblunt:
I'll re-run the jupyter notebooks over the weekend so you can see my additions!


henry-ngo:
Looks good overall. I am away from the office until Jan 2 since the work site is closed for the December shutdown period, so it's not super easy for me to run test plotting scripts. Can you show an output example in the Slack or maybe just explain the lines with `subplot2grid`? Looking at that `matplotlib` function, it seems like you are using this to place the zoomed in plot with the "main" plot or something?

sblunt:
example_orbit_figure

Here's an example of the new standard orbit plot. Let me know what comments you have! I added additional keywords to allow users to specify the "end date" of the Sep/PA panels and the color of the orbit tracks in those panels.

sblunt:
Yup! Done. I'll pull in the change now. Thanks all!

PR #91, opened 2019-01-11
colorbar flexibility

malenarice:
Add flexibility to the colorbar such that the user can select whether the colorbar is parameterized by either the epoch or an orbital element of their choice.

### DIFF
--- a/orbitize/results.py
+++ b/orbitize/results.py
@@ -332,9 +332,9 @@ class Results(object):
         elif cbar_param[0:3] in dict_of_indices:
             try:
                 object_id = np.int(cbar_param[3:])
-	    except ValueError:
-		object_id = 1
-                
+            except ValueError:
+                object_id = 1
+
             index = dict_of_indices[cbar_param[0:3]] + 6*(object_id-1)
         else:
             raise Exception('Invalid input; acceptable inputs include epochs, sma1, ecc1, inc1, aop1, pan1, tau1, sma2, ecc2, ...')
--- a/orbitize/results.py
+++ b/orbitize/results.py
@@ -330,7 +330,11 @@ class Results(object):
         if cbar_param == 'epochs':
             pass
         elif cbar_param[0:3] in dict_of_indices:
-            object_id = np.int(cbar_param[3:])
+            try:
+                object_id = np.int(cbar_param[3:])
+	        except ValueError:
+		        object_id = 1
+                
             index = dict_of_indices[cbar_param[0:3]] + 6*(object_id-1)
         else:
             raise Exception('Invalid input; acceptable inputs include epochs, sma1, ecc1, inc1, aop1, pan1, tau1, sma2, ecc2, ...')
--- a/orbitize/results.py
+++ b/orbitize/results.py
@@ -275,7 +275,9 @@ class Results(object):
                     object_to_plot=1, start_mjd=51544.,
                     num_orbits_to_plot=100, num_epochs_to_plot=100,
                     square_plot=True, show_colorbar=True, cmap=cmap, 
-                    sep_pa_color='lightgrey', sep_pa_end_year=2025.0):
+                    sep_pa_color='lightgrey', sep_pa_end_year=2025.0,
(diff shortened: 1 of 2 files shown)


semaphoreP:
@malenarice, feel free to message us on slack if you need help writing unit tests. Look at orbitize/tests/test_results.py in the `test_plot_orbits()` function. This is a test that will be run automatically each time someone commits the code. You could add additional lines to that tests that tests the optional color by orbital elements feature that you added. We are basically just checking to see that the code doesn't crash, since it is quite hard to test for "correctness" of the plot.

malenarice:
I just added a few lines to check the inputs for cbar_param and to raise an error when they are not the correct format (this is now in another commit on this branch). The `index` variable is only defined for the orbital element input parameters that are allowed by the function (sma1, ecc1, inc1, aop1, pan1, tau1, sma2, ecc2, ...), and is not created if the color bar is being set by 'epochs', which is the one other option that is currently accepted. Does this help to clarify what the earlier problem might have been?

I can add in a unit test, but I was wondering what exactly you had in mind -- i.e. would it be checking whether each parameter lies within a reasonable range, and raising an error otherwise?

henry-ngo:
Hi @malenarice 

The unit tests are just to ensure the code can run with your new options. Like @semaphoreP said, it's really hard to check if the Figure created is "right", so the current unit test just tries to generate a Figure with each of the combo of options and checks that the output is not `None`. I added a unit test for your branch in my local copy, which was how I found out that there was that bug. I think best practice is for you to be the sole person pushing commits to this branch and PR, so it would probably be best if you added the unit test (just a few lines). Then, you can either run the `test_results.py` unit testing file and/or let Travis CI automatically do the check after you commit your changes. 

I am not sure how to quote code lines yet, so I will try in the next message below!

henry-ngo:
Okay if I did this right, then this should be a permanlink to the `test_plot_orbits` method in `test_results.py`: https://github.com/sblunt/orbitize/blob/1c7fe798fe067c2d51baeaa118c5209d4d0f28d9/tests/test_results.py#L138-L150. You just have to follow the example and add another line like `Figure5 = ...` where you call the `plot_orbits` method with your new colourbar option (repeat for Figure6... if you have more than one new option). Don't forget to also add the `assert Figure5 is not None` which would throw an exception if Figure5 creation failed. Also, to keep consistent, add Figure5 to the tuple returned by this function.

Then, optionally, at the bottom in the `if __name__ == "__main__":` block, you can add lines to save the new figures. This is helpful for debugging locally since I can see the figures created and determine if it is what I want. I comment them out for the final commit though.

semaphoreP:
No reason why multiple people can't push to the same branch, except from a code ownership point of view (which I think doesn't really apply here?). Point of Github is to be collaborative!

henry-ngo:
@semaphoreP oh I must have misunderstood and/or misremember something you said a year ago then! I'll add a commit for the unit test shortly and @malenarice can add more tests if I am not trying all the cool features she implemented!

semaphoreP:
@henry-ngo Let's bring it up on the slack chat. So everyone is clear on PR and commit policies.

henry-ngo:
@semaphoreP  okay!

henry-ngo:
Okay, so now that we have the unit test, it looks like the Travis build failed. If you click on the "details" above, you can see the log and where it failed. Looks like line 333 of results.py:

```
        elif cbar_param[0:3] in dict_of_indices:
>           object_id = np.int(cbar_param[3:])
E           ValueError: invalid literal for int() with base 10: ''
```

semaphoreP:
@malenarice if you `git pull` and get Henry's update, you can run the test yourself on the local machine to reproduce and debug where it fails. Running `pytest` on the orbitize directory will run all of the tests, or you can specifically run `test_results.py` to just test Henry's update.

malenarice:
No that's okay, thank you for showing ways to make it more user-friendly! :)

henry-ngo:
Looks like it's all debugged now :D

Issue #95, opened 2019-01-25
warning message for OFTI results.plot_orbits


isabelangelo:
When running OFTI tests, the command:
s.results.plot_orbits(start_mjd=s.epochs[0])

yields several warning messages that look like (you can see this in the OFTI tutorial):
ErfaWarning: ERFA function "d2dtf" yielded 9625 of "dubious year (Note 5)" [astropy._erfa.core]

It looks like it comes from something going on under the hood in the Results class, but since the plot looks fine it might just be a warning we can ignore, in which case I can just put something in the code to hide the messages :)

sblunt:
Thanks for pointing this out, Izzy. I think it's just because many of the orbits we deal with have very long orbital periods, so the color bars on the orbit plots have to extend to like 3,000, which astropy thinks is a dubious year. It would be great if you could put something in the code to not print those dubious year warnings!

isabelangelo:
I put in a command to suppress the python "Warnings" category in plot_orbit() and uploaded it to the colorbar branch. I wasn't able to specifically target the ErfaWarnings specifically because the warning suppression function won't allow it, so this is more general. 

If it's too general and we think it might end up suppressing important warnings I can take it out though :)

sblunt:
See PR #96

semaphoreP:
Fixed in PR #96

PR #96, opened 2019-01-29
Filter astropy ERFA warnings

sblunt:
Hey @isabelangelo I think I figured it out. (see Issue #95)


isabelangelo:
oh nice, that makes sense that you can import the warning through astropy! it looks good to me :)

PR #98, opened 2019-03-18
Ofti speedups

isabelangelo:
This issue should be fixed now! I took Henry's suggestion and made it so sampler creates variables from copies of the original data table, so self.system.data_table will be constant.

Additionally, as we discussed, I made it so that OFTI has two distinct table attributes: self.input_table (same as self.system.data_table, identical to original file) and self.data_table (table that is used for OFTI analysis) and put in a print statement for when ra/dec are converted to sep/pa


sblunt:
Looks good to me! Happy to merge once Henry gives it the ok.

henry-ngo:
Looks great! :)

Issue #62, opened 2018-09-19
Custom tau start date


semaphoreP:
Somewhat related to #61. We should have an option for a user defined start date for tau (currently it's starting from MJD=0). There is possible evidence that having tau near the date of current observations helps with MCMC convergence.

semaphoreP:
So in discussions, we determined that tau is degenerate with period (as one might expect), so having tau start too far away from the actual observations creates weird covariances that are not Gaussian, and thus hard to sample. 

So, what should be the default? MJD=0 is too far back in time compared to most of our data. MJD=first astrometry data makes dataset dependent and possibly too confusing. My current suggestion is MJD=60000 (Feb 25, 2023), which is a nice round number close in time to our current data. Of course, we should have the option for the user to have a custom start date for tau.

henry-ngo:
How far away is "too far away"? Since most observations using orbitize will probably be from 2010 onwards, a date like 2023 is great if within ~10-20 years is close enough so that the default value will be good for as long as python is likely to be dominant in astronomy ;)

But if 10 years is already "too far away" then I'd prefer a date closer to today and no need to favour a nice round MJD.

sblunt:
I'm in favor of MJD=60000! I think we should use the same MJD convention for everything, including input astrometry. Because of that, I think we should make this a V1.0 issue. @semaphoreP would it be possible to do an MCMC run with beta Pic and MJD= 60000 to make sure it's not "too far away?"

sblunt:
Oh wait I'm stupid never mind about the v1.0 issue, I didn't think that comment through. But I still agree with my suggestion to do an MCMC run with MJD=60000!

semaphoreP:
Implemented in PR #105

Issue #61, opened 2018-09-19
Helper function to work with tau


semaphoreP:
Tau is great for sampling, but somethings annoying to work with. Instead of having each person re-figure out tau every time, it'd be great to have some helper functions in orbitize. I was thinking of the following

- Convert tau to T0 (actual MJD date of periastron passage). Optionally, T0 can be the first periastron passage after a certain date.
- Convert tau to a different tau (we base tau off of MJD=0, but having the flexibility of switching to a different date is also good).

I think the first one is the most important, but the second one could also be useful while we are attacking this problem.

semaphoreP:
Fixed in PR #105

PR #105, opened 2019-05-03
Tau utilities and customization

semaphoreP:
This PR addresses issues #61 and #62. The key changes are

- utilities in manipulate tau (change reference epoch, and get date of periastron) using the basis.py module
- system/driver now can set the tau reference epoch as an init argument
- default reference epoch is Jan 1, 2020 (MJD=58849) instead of the old MJD = 0
- Results class now saves tau reference epoch to the header and can handle the reference epoch internally so user can still pass in MJD dates

This PR modifies the Results and OFTI classes, as well lots of other miscellaneous bits of code, so it would be good to verify I didn't mess up anyone's code. Another open question is whether we want to call it basis.py or something else -- my hope in the future is that if we have different sets of orbital elements, we could include them in this module.

### DIFF
--- a/orbitize/basis.py
+++ b/orbitize/basis.py
@@ -1,0 +1,65 @@
+import numpy as np
+import astropy.units as u
+
+def tau_to_t0(tau, ref_epoch, period, after_date=None):
+    """
+    Convert tau (epoch of periastron in fractional orbital period after ref epoch) to
+    T0 (date in days, usually MJD, but works with whatever system ref_epoch is given in)
+
+    Args:
+        tau (float or np.array): value of tau to convert
+        ref_epoch (float or np.array): date (in days, typically MJD) that tau is defined relative to
+        period (float or np.array): period (in years) that tau is noralized with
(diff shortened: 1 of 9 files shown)


henry-ngo:
It looks pretty good to me. I only have a few suggestions for comments and docstring. Is it okay if I push a commit to your branch with these minor things?

sblunt:
It looks good to me! I appreciate the backwards compatibility and I like the basis module. Having the choice of tau actually written down in the results file is a really good idea.

semaphoreP:
@henry-ngo please go for it. I realize I didn't write the doc page I said I was going to write also.

henry-ngo:
Okay, I pushed two commits yesterday evening! Looks like they appear right below my comment rather than after your reply (perhaps it goes by commit time instead of push time?) Anyways, I think that covers all of the docstrings and the rest looks good!

isabelangelo:
Looks good to me! I think this will help with clarity about how tau is calculated/defined, which is great!

Issue #108, opened 2019-06-05
Input Table with RA/DEC offsets not working/Only works with SEP/PA


JunellieGnzlz:
I've been working with orbitize now for a couple of weeks (great tool!), and I wanted to raise an issue that came up when inputting the data to orbitize. When trying to read an input table with RA and DEC offsets, a couple of things happened:

1. After reading in the table with RA and DEC offsets successfully with the orbitize.driver.Driver, it prints out "Converting ra/dec data points in data_table to sep/pa. Original data are stored in input_table." But when you print out the data table with myDriver.system.data_table, it prints out the original RA and DEC offsets, instead of printing out the table with the converted values to sep/pa as stated in the statement. There's no sign that it was actually converted.

2. When I try to run the run_sampler for OFTI with the input table containing RA/DEC, it runs forever. It actually doesn't stop running. It doesn't even show the "x/1000 orbits found" progress statement. 

NOTE: When I tried converting the values myself from RA/DEC to SEP/PA using the equations in the orbitize source code and inputting it as a SEP/PA table, it ran smoothly. No issues there, which makes me believe there is some sort of bug when inputting the RA/DEC tables where it doesn't convert the values and doesn't let the sampler run as it should.

henry-ngo:
Hi @JunellieGnzlz ! Thanks for the report. I'll check this out. I just recently did something similar but I read the input table directly with the `read_input` module. So it might be something in `Driver` that is causing the issue(s).

sblunt:
Thanks for using orbitize, @JunellieGnzlz! And thanks for handling this bug, @henry-ngo!

henry-ngo:
Hi @sblunt , @logan-pearce and @isabelangelo:

I think I found the cause of the issue! Context, in case you find a reminder useful: We recently changed the behaviour of the OFTI sampler when it is given ra/dec as input (as @JunellieGnzlz has done) so that we save the actual input as an attribute of the `sampler` object (i.e. `sampler.input_table`) instead of simply overwriting the old line in `data_table`. 

[...]

PR #109, opened 2019-06-07
Fix radec/seppa conversion with OFTI sampler

henry-ngo:
This is a fix for Issue #108 

As discussed, I remove `sampler.data_table` and `sampler.input_table` in favour of `system.data_table` and `system.input_table`. I moved the actual radec/seppa conversion calculation to be done within the `system` object, as I wanted it to change `system.data_table` in place while also updating the arrays `system.radec` and `system.seppa` used to keep track of which entries are in which format. I believe these arrays are necessary in additional to the `quant_type` column of `data_table` because it sorts them by body number. 

Right now, these conversions are done for one body at a time since we don't do multi-planet yet. It seems like the OFTI sampler was already hard-coded for just 1 planet so I kept it that way to keep the scope of this PR only on #108. It can be made to accommodate multiple planets when we make that bigger update!

### DIFF
--- a/orbitize/sampler.py
+++ b/orbitize/sampler.py
@@ -88,11 +88,11 @@ class OFTI(Sampler):
     Args:
         like (string): name of likelihood function in ``lnlike.py``
         system (system.System): ``system.System`` object
-        custom_lnlike (func): ability to include an addition custom likelihood function in the fit. 
+        custom_lnlike (func): ability to include an addition custom likelihood function in the fit.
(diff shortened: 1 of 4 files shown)


sblunt:
Looks good to me! Thanks again @henry-ngo! I agree that the function name is long but I like that it's descriptive.

isabelangelo:
I think this is definitely an improvement, thanks Henry!

Issue #111, opened 2019-06-26
Error reporting for read_input


sblunt:
After spending all day at the WFIRST data challenge with folks using orbitize, I've seen a few issues come up several times. I think they could be easily averted with more descriptive errors. The two bugs I saw several times today happened when:

- the path to the input csv was incorrect
- the input epochs were not in MJD

It would be great to add some error messages for these specific cases.

sblunt:
Looks great. Thanks @henry-ngo. I purposefully didn't assign you to this issue because I didn't want to create more work for you but you were too quick!! :D I just added a message notifying the user that JD was changed to MJD. Let me know what you think of this addition, and I can pull the changes in.

henry-ngo:
haha, I figure I should take ownership of read_input related issues :) Everything looks good!

sblunt:
Addressed in PR #112

PR #112, opened 2019-06-28
Added two exception messages 

henry-ngo:
to let the user know there was a problem finding/reading the file and if the epoch was not given as MJD (or JD).

In response to Issue #111.


henry-ngo:
(let me know if you think more detailed checks would be useful!)

Issue #117, opened 2019-07-04
Default behavior for restrict_angle_ranges


jruffio:
For its default behavior, it looks like the intent of the code is to set the restrict_angle_ranges  flag based on the dimensions of the input data (although it is not described in the docstring). However, I think this is currently not working:

The line 

> if (len(radec_indices) + len(seppa_indices) == len(self.data_table)) and (restrict_angle_ranges is None):

should be replaced by

> if (len(radec_indices[0]) + len(seppa_indices[0]) == len(self.data_table)) and (restrict_angle_ranges is None):

sblunt:
Thanks JB. I decided to remove these lines of code after all. I'll let the user/sampler object decide when to restrict the angle ranges.

sblunt:
See PR #126

sblunt:
We actually ended up addressing this issue in PR #131.

sblunt:
Re-removed in PR #126

Issue #113, opened 2019-07-02
OFTI conversion of errors from RA/Dec to Sep/PA is incorrect


semaphoreP:
The `radec2seppa` function can't be used to directly convert errors. See below (~line 126 in sampler.py). 
```
self.sep_err[i], self.pa_err[i] = radec2seppa(
                self.sep_err[i], self.pa_err[i]
            )
```
I think the error in separation should be the average error in RA and Dec, and the error in pa is sep_err/sep.

sblunt:
Addressed in PR #126

Issue #103, opened 2019-04-02
``OFTI`` adds likelihood probability to ``results.py``, while ``MCMC`` adds posterior probability


sblunt:
Just putting this minor bug here so I remind myself to address it. Currently, the OFTI sampler is adding likelihood probabilities (i.e. not multiplied by prior probabilities) to the ``Results`` object, and the MCMC sampler is adding posterior probabilities (i.e. with prior probabilities included). Not a big deal, and doesn't affect performance at all, but they should behave consistently.

henry-ngo:
Which one would be more useful? Or any reason to keep both? My first thought is that posteriors are preferred since that's what determines which samples get chosen in the first place?

semaphoreP:
Does the OFTI samples inherently have the prior probabilities in them, because they get drawn, weighted by the prior probability?

sblunt:
Yes, prior probability is incorporated in how frequently various orbits are drawn, but the number OFTI currently returns is just -0.5*chi2. I guess it's just a matter of semantics, but that makes more sense to me as a likelihood. To calculate an a posteriori probability for a given OFTI orbit, you'd have to additionally incorporate the priors.

Maybe we could keep likelihood probability and prior probability separately? Having both won't increase the size of the save file toooo much, so perhaps that's justified. I agree that having the posterior probability is useful, but likelihood probability is nice too in case people want to report statistics about chi2, for example.

sblunt:
The problem I'm facing now is that I want to look at chi2 for orbits run by OFTI and MCMC, and currently that requires treating OFTI and MCMC outputs differently.

sblunt:
Does that make sense?

semaphoreP:
Yes, I understand now! Basically our posteriors derived from the samples are consistent, but the reported probabilities correspond to different things. I agree that sounds like a good thing to keep straight. I think keeping both, or just keeping posterior probability are sensible solutions. In my experience, it doesn't take that much CPU time to recompute likelihood for chi2 if need be. I think either solution is good!

henry-ngo:
Yeah, that makes a lot of sense! Thanks for the explanation / example of a use of the likelihood. In that case, I think it makes sense to have both likelihood and posterior (ptemcee/emcee provides both anyways). With knowledge of the prior, it's possible to get the posterior from just the likelihood or vice-versa, but more user-friendly to just have both!

henry-ngo:
[...]

PR #125, opened 2019-07-10
Kepler mass redef

sblunt:
Medium-level API update. Under the new convention, `orbitize.system.System` objects are initialized either with a `mtot` parameter alone or with a `m0` and `m1` parameter. This allows for easier handling of dynamical mass fits with RV data + astrometry. Also, `orbitize.results.Results.save_results()` now saves the list of parameter labels in addition to the posterior, the loglikelihood, and the tau reference epoch.

PR #126, opened 2019-07-11
Fix three bugs

sblunt:
Issues #113, #103, #117


sblunt:
Note that this PR builds on PR #125.

semaphoreP:
Does `pa_err = sep_err/sep` return in radians, whereas PA is defined in degrees?

henry-ngo:
Not sure if I should comment on #103 or here.  At least this line will cause a automatically link to appear in #103, right? Hopefully that github magic works lol.

It looks like you said you want to only return lnlike for both OFTI and MCMC. Also looks like the changes rename the lnpost entries in Results object to lnlike. 

But I believe the MCMC sampler will always return lnpost, not lnlike. Maybe I missed a change to sampler.py but since we are passing prior info to the MCMC sampler, all samples chosen by ptemcee will already incorporate prior info so the distribution we get in the end is actually a sample of the posterior, not the likelihood? 

I think most MCMC users would expect to get the posteriors, not the likelihoods. So I think the current implementation is fine but the naming convention could be misleading. 

But if I misunderstood the changes then please let me know!

sblunt:
Yes, thanks for that good catch @semaphoreP. I just fixed it.

@henry-ngo, yes emcee by default returns log posterior probability, but I subtracted off the log prior probability so it should give likelihood.

henry-ngo:
@sblunt oh okay, I see the changes now!

I don't think we need to subtract off the prior when using the Parallel Tempering MCMC sampler because it returns likelihood and prior separately, since it needs to calculate them separately for the temperatures. the `lnlikelihood` property of the ptemcee sampler should get you this! We should just grab this attribute instead.

However, for the EnsembleSampler, we don't get likelihood separately. Just thinking of an edge case for that: if the prior returns -np.inf, then this would make the likelihood be +np.inf. Although we should never have this since a location that is outside the prior would not be sampled by emcee.... Unless, the walker inadvertently started outside of the prior space and couldn't find its way back in. Then the posterior returned should be -np.inf. The prior calculated is also -np.inf. The likelihood would be `nan` here, since I think -np.inf - (-np.inf) = nan? This isn't ideal since it's a little confusing and the user can't get back the posterior by adding the prior back on. 

The ideal thing to get here would be the actual likelihood value. Perhaps we can do a check for nans in this array and then calculate the likelihood again for the steps that have this problem. This can be computationally costly and if these are outside of the prior, then maybe they are not useful anyways and the user might not know this until it spends a bunch of computational time calculating useless values. So maybe returning nans (to distinguish from -np.inf likelihoods) with a warning? Or setting nan to -np.inf with a warning?

semaphoreP:
Henry, these are good points. It does feel a little weird to me to do Bayesian orbit fitting but only save likelihoods and not posterior probabilities. The latter does seem like the better Bayesian criteria, although frequentist metrics to assess goodness of fit using the likelihood are sometimes useful. Perhaps we should try to save both? Did we have a discussion already on whether to save both (I forget)?

henry-ngo:
@semaphoreP : I think we talked about this a bit here and maybe at the hackathon too? @sblunt said she made a decision in the thread for Issue #103

sblunt:
That's a good point @henry-ngo. I think returning a warning would be good. In general, it seems like a good idea to warn the user if the chains are starting outside the prior space. 

Yes, @semaphoreP here's my comment on Issue #103. Open to other ways of dealing with it though.

"I decided to just return the loglikelihood, rather than both it and the logprobability, since the results object code was getting messy with lots of try-except cases for backwards-compatibility. Users should be able to generate lnprobs without too much difficulty, and the code Junellie is working on (Issue #118) will make this more user-friendly too."

(ps. I think I treated ptemcee and emcee outputs correctly, but maybe worth checking since y'all are the MCMC experts?)

semaphoreP:
Looks right to me. Maybe best thing to do is write a test to confirm emcee stores the likelihood (i.e., compare the lnlike in results with straight up computing the lnlike using the lnlike function)?

henry-ngo:
Hi @sblunt 

On returning the right property from PT MCMC vs. EnsembleSampler MCMC: I think this should be a permalink to the lines of code that I woudl like to discuss: https://github.com/sblunt/orbitize/blob/4c83586698a80e0d65389a91e02355a7408189fd/orbitize/sampler.py#L490

I think Lines 492 and 493, within the `if` block when running PT should have `sampler.logprobability` changed to `sampler.loglikelihood` so that we directly get the likelihood from the sampler, instead of computing it ourselves as we do now in Lines 499-500. With this change, you can move the current prior subtraction on Line 500 to be within the `else` block. 

What you have now will still calculate the correct result. However, in the PT case, I would prefer to directly get the likelihood where possible. It will avoid potential errors where the priors are not well behaved values as I wrote above. For the emcee sampler, we can't get around it so just raising a warning should be fine.

But if you have a different reason for choosing logprobability there in Lines 492-493, I'd love to hear it!

sblunt:
Hey all, I fixed the point Henry pointed out and added the test Jason suggested. This branch is ready to merge now! I'm going to pull in the changes then release a new version.

PR #131, opened 2019-08-05
omega bounds should always be [0,2pi]

jruffio:


semaphoreP:
JB, can you also modify the docstring for the restrict_angle_ranges so that it indicates we are only changing \Omega. 

You mentioned before that the logic for restrict_angle_ranges doens't always work... has that been fixed?

semaphoreP:
Can you also put in a print statement when the code automatically sets restrict_angle_range=True so that the user is informed that we are changing the prior of \Omega to 0 and 180 deg?

semaphoreP:
@sblunt Are we ok with Omega being fit between 0 and pi by default when there is no RV data? I think that was our original intent but orbitize has not been behaving like that due to a bug. Just wanted to check.

sblunt:
Hey all, apologies for my radio silence the last few days. Yes, the behavior you suggest is fine with me, Jason.

sblunt:
Thanks for finding and fixing this bug, @jruffio !

PR #133, opened 2019-08-17
Parallel processing for OFTI

sblunt:
Have a look and let us know what you think, all!


semaphoreP:
Here's some relatively minor comments. Overall looks good!

Is `run_sampler_base` supposed to be a hidden/private function? If so, I normally name it starting with a `_` (e.g., `_run_sampler_base`). 

Would using `mp.Pool()` be a cleaner implementation than spawning a thread pool manually using `Process()`. This would remove all of the process handling code including the need to pass Value around I think. Happy to discuss this more in slack if you would like. 

We should have a smarter default for `num_cores` other than 8. If you pass `None` to multiprocessing, it will automatically use all of the cores. Otherwise, we can use `mp.cpu_count()` to get the number of cores on a computer, and adjust accordingly from there (e.g., `ceil(8, mp.cpu_count())`)

henry-ngo:
I agree with Jason on the naming of `run_sampler_base`. Also, if it is indeed only meant to be called by `run_sampler`, I think we should make this explicit in the description of the function in the docstring. We might even consider naming it something else entirely so that users aren't confused about which one is the "driver" function.

I also agree that `num_cores` should not default to `8`. I think the default to max cores is a good idea, or default to 1 core (I think this the MCMC default?)

Finally, one new comment/question: What happens if a user calls this with `num_cores=1`. Is it exactly the same as before? Does it still create the MP objects and slow things down? Ideally, calling it with `num_cores=1` should skip all of the steps needed for multiprocessing.

vighnesh-nagpal:
Hey Henry, Jason!

Thanks for the suggestions! Based on them, I've made a few changes to the code.

- num_cores now defaults to max cores.
- if num_cores=1, then the code executed is the same as the original run_sampler so none of the mp stuff is used
- I've renamed run_sampler_base to _run_sampler_base, although I do agree that naming it something else entirely might be more clear. Do you guys have any ideas for alternate names that would be better?
- The documentation for _run_sampler_base now mentions that the function is meant to be called by run_sampler

I initially tried using mp.pool, but then I had issues with getting the worker processes to converge. Would successfully using mp.pool be a significantly faster implementation?

PR #137, opened 2019-09-16
Mikkola

devincody:
implementation of mikkola solver in c.


devincody:
Changed all my tabs to spaces to be more consistent. That's why there are so many line changes

PR #146, opened 2019-10-18
Rv changes

Rob685:
MCMC rv capability


sblunt:
Hey @Rob685! Nice job getting this branch ready to go. There were a few merge conflicts with master that I fixed up (check out the most recent commit to this branch, which will show my changes). I didn't change anything major, but you should have another look over to make sure your own tests are still passing and that the code still looks good to you. Once that's done, and the tests are [passing on travis ](https://travis-ci.org/sblunt/orbitize/builds/600425598?utm_source=github_status&utm_medium=notification), I'll do a more thorough look through.

sblunt:
Ok it looks like `test_mcmc_rv.py` is failing. Can you try to figure out what's going wrong with the test? Does it work on your computer?

lhirsch238:
Rob and I have discussed an error that is happening in the rv-changes branch, specifically when the mass and parallax are fixed rather than included as free parameters with Gaussian priors. In the case of fixed parameters, the chains object stores p-2 parameters (not including total/stellar mass or parallax), and the likelihood function does not know how to properly interpret this truncated array. 

We need to take a look at master to see how to solve this bug, but I think that's where it's coming from.

semaphoreP:
For MCMC sampler, it's `logl()` function first calls a function called `self._fill_in_fixed_params()` so that the `system.compute_model()` parameter has all of the values. That fill in function should be pretty general, so I would have expected it to work, but that's where it should be happening.

semaphoreP:
Hi Rob,

How's the progress going? We should think about merging in the MCMC RV update into master if it's working. You can continue to work on the branch to do OFTI stuff, but it might be good to break up the merges so this PR isn't open for too long.

Rob685:
Hi Jason,

Thanks for the nudge! Yes the MCMC RV is working now and I'm currently working on troubleshooting the Travis tests so we can finish this up. Sorry about the delay.

sblunt:
Hey @Rob685! I took a more careful look through the RV tutorial, and I love it! It's clear and complete. I have a few minor requests, and one more major request that we might just want to punt to a later time:

1. Please add a citation to the new data csv file. The other data csv files just have an extra comment line that give the reference-- something simple like that is totally fine.

2. I anticipate a few FAQs about the RV implementation which would be good to answer at the beginning of the tutorial; would you mind adding a sentence or two saying that the RV+astrometry implementation currently works for MCMC (ie not OFTI) and does not account for potential gamma/jitter differences between instruments?

3. - Ideally, the tutorial should be able to run on anyone’s computer without any modifications. Would it be possible to change some of the hard-coded paths in the tutorial to relative paths so they can run without modifications?

-- Larger request --

4. It would be great to add a panel to the orbit plots that visualizes the RV timeseries for fits that have RV measurements. Is this something you and @lhirsch238 would be interested in working on? If not, I still think we should merge this branch, but we should remember to make it an issue to work on at a later date.

sblunt:
One more quick comment: is it necessary to keep the changes in `OFTI.__init__()` now that we are not requiring that OFTI run on RV+astrometry datasets? I think it might be good to instead just print an error message if a user tries to run OFTI on an RV+astrometry dataset (basically saying "go run MCMC instead").

PR #147, opened 2019-11-09
Default corner plot now plots angles in degrees

sblunt:
Partially in response to the referee, partially in response to everyone else who has (rightfully) complained about this. orbitize! output is still given in radians.


sblunt:
Yes, definitely. Thanks for the note. Should be fixed in latest commit. Lmk if you have any other suggestions!

PR #151, opened 2020-01-19
Plot RA with positive to the left as is convention

semaphoreP:
I just noticed that the orbit plots have RA in the wrong direction. Positive RA should go to the left, not the right. I added one line of code to invert the x-axis. I edited this online, so definitely wait for the tests to pass first to make sure I didn't make any typos.

### DIFF
--- a/orbitize/results.py
+++ b/orbitize/results.py
@@ -419,6 +419,7 @@ class Results(object):
             ax.set_ylabel('$\Delta$Dec [mas]')
             ax.locator_params(axis='x', nbins=6)
             ax.locator_params(axis='y', nbins=6)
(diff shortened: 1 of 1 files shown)


semaphoreP:
Hmm, yeah. I'm not sure what is happening with travis. Did numpy or astropy update with some breaking changes?

Good question on whether to include an optional argument. I currently can't think of a case in which you would want to flip it, and we do return the `figure` to the user, so if they really want to they can. If anyone chimes in with a use case for flipping it, then we should add it. Otherwise, I think if it's niche enough, we can leave it to the user to modify.

henry-ngo:
sounds good to me!

semaphoreP:
horray! this works now with the new python3 requirements.

Issue #152, opened 2020-01-21
Tests failing on master branch


aryadas98:
Tests on the latest master branch are failing. From the root directory of the project if I run `pytest` these errors come up:

```
======================================= test session starts ========================================
platform linux -- Python 3.6.7, pytest-5.3.4, py-1.8.1, pluggy-0.13.1
rootdir: /home/arya/orbitize
collected 18 items / 6 errors / 12 selected

============================================== ERRORS ==============================================
_______________________________ ERROR collecting tests/test_OFTI.py ________________________________
venv/lib/python3.6/site-packages/numpy/core/function_base.py:117: in linspace
    num = operator.index(num)
E   TypeError: 'float' object cannot be interpreted as an integer

During handling of the above exception, another exception occurred:
tests/test_OFTI.py:10: in <module>
    import orbitize.sampler as sampler
orbitize/sampler.py:19: in <module>
    import orbitize.results
orbitize/results.py:27: in <module>
    cmap(np.linspace(0.0, 0.7, 1000.))
<__array_function__ internals>:6: in linspace
    ???
venv/lib/python3.6/site-packages/numpy/core/function_base.py:121: in linspace
    .format(type(num)))
E   TypeError: object of type <class 'float'> cannot be safely interpreted as an integer.
```

The rest of the errors are caused by the same line `cmap(np.linspace(0.0, 0.7, 1000.))`. The problem is that the last argument 1000 should be an integer. However even after removing the period, the tests are failing.

After removing period:
```
======================================= test session starts ========================================
platform linux -- Python 3.6.7, pytest-5.3.4, py-1.8.1, pluggy-0.13.1
rootdir: /home/arya/orbitize
collected 38 items

tests/test_OFTI.py ...                                                                       [  7%]
tests/test_api.py FF...                                                                      [ 21%]
tests/test_conversions.py ...                                                                [ 28%]
tests/test_driver.py FF.                                                                     [ 36%]
tests/test_kepler_solver.py .........                                                        [ 60%]
tests/test_mcmc.py F.                                                                        [ 65%]
tests/test_priors.py ..                                                                      [ 71%]
tests/test_read_input.py FFFF                                                                [ 81%]
tests/test_results.py ....                                                                   [ 92%]
tests/test_system.py ...                                                                     [100%]

============================================= FAILURES =============================================
________________________________________ test_compute_model ________________________________________

    def test_compute_model():
        """
        Test basic functionality of ``System.compute_model()``
        """
[...]

PR #154, opened 2020-01-22
Update results.py

henry-ngo:
changed call to np.linspace() to have an integer as the `num` argument instead of casting a float, which is deprecated as of numpy v1.12

addresses one part of #152


sblunt:
Thanks for making this PR, @henry-ngo! I forgot that you had already done this, and I made the change myself. Version 1.8.0 has this change implemented. I'm sorry I didn't pull in this branch.

henry-ngo:
no problem! this was before the larger discussion on python2 support otherwise it would be my smallest PR ever (1 character change!)

PR #161, opened 2020-03-09
Multiplanet support

semaphoreP:
This is both the multiplanet support, and changes to the packaging so that the .csv data files are included as part of orbitize! release. 

For multi-planet support, system handles multiple planets and computes the perturbations on the orbits due to the other planets (planet-planet effects ignored), OFTI/MCMC sampler handles multiple planets, and result plots can handle multiple planets. There is also a new tutorial on multiplanet support. Let me know if there's any comments!


semaphoreP:
Thanks! I just added some changes to address the first 3 comments. Let me know how they look.

Regarding tests, I don't know of any established fits to compare to. We could simulate data with orbitize! and do a self-consistency check? I have an uncommitted multiplanet test that tests `system.compute_model()` against the REBOUND N-body simulator for 1 and 2 body cases with and without mass. The 2 planet with masses test passes as long as I relax the precision of the test to 1e-4 precision. Maybe this is good enough? We would need to install REBOUND on travis for this test. 

Let me know which of those options makes the most sense!

sblunt:
Thanks! I think the first three points look good.
Simulating data with orbitize and doing a self-consistency check sounds good to me. I mostly just want to make sure we have a unit test that actually uses this combination of options. I don't think we need to do a full comparison with rebound in the unit tests.

semaphoreP:
Ok, I did a end-to-end test of multiplanet functionality, and a unit test of the compute_model with multiple planets. Multiplanet is a little slow (possibly an area to investigate) so I ran a very short MCMC.

sblunt:
Cool! Ready to pull in? I think there's a lingering merge conflict.

semaphoreP:
Oops, I meant to fix that. Should be fixed now.

PR #165, opened 2020-04-15
Fix Multiplanet Test

semaphoreP:
I removed eccentricity from the MCMC test for multiplanet. It should be easier for MCMC to converge without it now. Couldn't get the numpy.seed() working so that MCMC is predictable, but I ran it multiple times and now the masses change by < 10%.


sblunt:
Looks good, thanks!

PR #170, opened 2020-05-05
improved mtot calculation for multiplanet

semaphoreP:
This fixes an issue where we weren't properly handling what should be the total mass in multi-planet systems (affects the orbital period). I checked against N-body integrators, and we should use the sum of all masses with r <= distance of the planet when computing mtot. You can think of this as an application of Gauss's law where the force on the planet is equal to the force if all the mass within a sphere is at the center of a sphere. 

I think this is right at least...

### DIFF
--- a/orbitize/system.py
+++ b/orbitize/system.py
@@ -266,7 +266,7 @@ class System(object):
 
             # vz_i is the ith companion radial velocity
             if self.fit_secondary_mass:
-                vz0 = vz_i*-(mass/m0)  # calculating stellar velocity due to ith companion
+                vz0 = vz_i*-(mass/mtot)  # calculating stellar velocity due to ith companion
                 total_rv0 = total_rv0 + vz0  # Adding stellar velocity and gamma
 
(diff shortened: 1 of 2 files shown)


semaphoreP:
Thinking about this more, I am wondering if our use of `-(mass/m0)` is correct, or whether it should be `-(mass/mtot)`. I will do a bit of testing on this before I merge these changes. I recall we discussed this before, but I forget the reasoning.

semaphoreP:
Ok, checked against REBOUND and checked against my classical mechanics textbook. I believe when going from the 1-body coordinate system to computing the coordinates of one of the two bodies relative to the center of mass, we need to use mass/mtot and not mass/m0. I made the adjustments now. It should only matter if `sigma_data/data ~= mass/mtot` which I think for stellar RV and astrometry of imaged systems, this should not matter yet.

sblunt:
Cool, looks good! Feel free to merge whenever you want.

PR #180, opened 2020-07-15
Degrees to radians and epoch to tau

### LINKED ISSUE #179: modifying mcmc positions docs refers to Epp instead of tau
In Code block [8] of [the tutorial](http://orbitize.info/en/latest/tutorials/Modifying_MCMC_initial_positions.html#3.1)-Generate-your-own-initial-positions), we reference what appears to be an epoch of periastron passage (in days) instead of tau from Blunt et al. 2017. This could be misleading since we need tau to initialize the walkers positions.

rferrerc:
Fixes #179 

I converted the angle variables from degrees to radians, and converted the epoch to tau using the values for the period in Table 1 of Blunt et al. 2017, AJ, 153, 229.


sblunt:
@rferrerc this looks ready to go once you've made the change Jason suggested.

Issue #163, opened 2020-03-25
MCMC examine_chains() should have a transparency option


semaphoreP:
We plotting more than 10 chains, or very long chains, the walker plots generated by `mcmc.examine_chains()` are hard to read if we don't have transparency enabled. Example below:

![image](https://user-images.githubusercontent.com/2245740/77560736-13c60a00-6e7b-11ea-9516-ec1f6da282ee.png)

Transparency will help us better have a sense of the density of walkers, which otherwise we can't tell. We can make it an optional user argument to tune the level of transparency.

sblunt:
Addressed in PR #183

PR #183, opened 2020-07-23
Added transparency to MCMC examine_chains function

sofiacovarrubias:
Optional transparency variable where default is 1 (100% opacity).


semaphoreP:
Hi Sofia, overall looks good! Just a couple quick things:

1. Can you rerun the MCMC_tutorial jupyter notebook so that all the cells return good values and not error messages? The outputs of the cells get used in our docs, so we don't want error messages.

2. Can you add a line to the docstring at line 717 that describes the transparency parameter? You can use the current ones as a template. 

Thanks!

sblunt:
Addresses issue #163

Issue #182, opened 2020-07-22
tau_ref_epoch inconsistency


semaphoreP:
The default value for `tau_ref_epoch` in `kepler.calc_orbit` is 0, while in the rest of the package, `tau_ref_epoch` is at `58849`. This is technically a breaking change in case other people are calling calc_orbit assuming tau_ref_epoch=0 works for them, but that subset of orbitize users may just be me.. (but nevertheless, we should document the change in the release notes).

sblunt:
Addressed in PR #187

PR #187, opened 2020-09-04
 warning for tau_ref_epoch change in kepler.calc_orbit

semaphoreP:
Addresses issue #182. After changing the hard coded tau_ref_epoch to be in line with default value in system.py, I added a new optional flag to `kepler.calc_orbit` that automatically throws a warning each time it's called, unless the flag is set to False. I've set the flag to False for every internal orbitize call to `calc_orbit` so users just using standard orbitize! API will not get this warning. However, users calling it from their own scripts will get this warning, and will need to explicitly turn it off themselves. 

How does that sound for a solution to transition out of the default tau_ref_epoch being 0 in calc_orbit?


sblunt (approved):
Looks great! This is a very nice way to resolve this annoying issue. Thanks @semaphoreP!

sblunt:
Addresses Issue #182

PR #189, opened 2020-11-11
Radvel utils

sblunt:
- fixes mass/mtot bug that was causing small errors in RV calculation for joint fits
- add a Radvel->projected separation posterior utility function & accompanying tests


orbitize/radvel_utils/compute_sep.py
```diff
@@ -0,0 +1,83 @@
...
+
+    Args:
+        df (pd.DataFrame): Radvel-computed posterior (in any orbital basis)
+        epochs (np.array of astropy.time.Time): epochs at which to compute 
+            separations
+        basis (str): basis string of input posterior (see 
+            radvel.basis.BASIS_NAMES` for the full list of possibilities). 
+        mtot (float): median of total mass distribution (assumed Gaussian).
+        mtot_err (float): 1sigma error of total mass distribution 
+            (assumed Gaussian).
```
lhirsch238, line 23: I think my only comment is about mtot vs. m0. For planets this doesn't matter, but will we run into issues if we're doing binary fits? We usually only have a gaussian prior on m0, not mtot for a binary system. Should the code be able to convert K and mstar to mtot more generally?

sblunt: Definitely, great point. I'll amend that.

lhirsch238 (approved):

semaphoreP (approved):
looks good!

Issue #191, opened 2020-11-12
multiplanet interactions should not include outer planet perturbing inner planet


semaphoreP:
Pointed out by Tim Brandt and Mirek Brandt that the inner planet shouldn't be perturbed by the outer planet to get the best agreement between orbitize! and REBOUND. Just some logic needs to be added to the system class.

sblunt:
Addressed in PR #192

PR #192, opened 2020-11-17
account for inner planets only in perturbations

semaphoreP:
Only inner planets perturb the visual orbits of outer planets. Outer planets are not accounted in inner planet orbits. I checked against REBOUND and this is indeed a better approximation (no test yet, unless we want to formally require REBOUND in our CI). 

Fixes issue #191

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ before_install:
 
 install:
     - conda install --yes python=$TRAVIS_PYTHON_VERSION pip numpy scipy cython matplotlib
-    - pip install pytest pytest-cov coverage coveralls corner deprecation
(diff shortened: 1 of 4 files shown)


sblunt:
The code looks good to me! I think it would be worth having an explicit rebound comparison test here. I don't mind adding rebound to the requirements, especially since we'll have to do it soon anyway.

semaphoreP:
I just realized REBOUND doesn't work on windows, so I might look try just put it on our Travis CI and use it for that one test specifically.

sblunt:
Looks great! Thanks for adding that test!

sblunt (approved):

PR #196, opened 2020-12-01
Community agreement

sblunt:
Add community agreement to `contributor_guidelines.md` and index page of docs


semaphoreP (approved):

semaphoreP:
It looks like there's an unrelated astropy deprecation issue that's causing builds to fail.

PR #198, opened 2020-12-15
2 wrongs made a right.

semaphoreP:
oops, had two bugs, which cancelled each other out and was causing all the tests to pass, but the code to not technically be correct in all use cases.


semaphoreP:
this should fix a bug introduced in #192

sblunt:
Looks good, I like the added multiplanet vs rebound test. I think I'll restructure a bit of this code (`compute_module`) when I finish the Hipparcos module, as we discussed, but it's working so let's pull it in.

sblunt (approved):

PR #199, opened 2020-12-16
compute_sep should also return mp

sblunt:


lhirsch238 (approved):
Yep, I agree!

semaphoreP (approved):
Looks good! Just one quick question, in orbitize with fit_secondary_mass, we usually use the basis order (sma, ecc, inc, aop, pan, tau, plx, m1, m0). Here we end it with (mtot, m1). I'm not sure how this function is used, but I was wondering if it makes sense to keep it consistent.

sblunt:
Good call, I'll make that change.

PR #204, opened 2021-01-25
Opencl

sblunt:


sblunt:
To dos (@devincody):
- add a global flag to kepler.py that sets which GPU solver to use (incl. None). 
- re-run tests.
- add documentation to kepler.py that briefly describes the two GPU solvers and tells users how to use them.

(same list applies to PR #203)

sblunt:
Hey @devincody I merged the latest commits from `develop` into this branch, and also added a keyword to calc_orbit to expose the openCL solver to users. 

I'm trying to get the OpenCL solver to work on my machine, but getting a segfault. Is that because I don't have the correct hardware or because something's up with the code? It would be good to check that a user has the right hardware if the former. 

I'll leave this branch for now, but when you have a chance it would be great to consolidate the two PRs and decide what to do with this code moving forward. Thanks for all of your work on this! :) <3

devincody:
Pycuda https://github.com/sblunt/orbitize/pull/203 is the GPU implementation branch that will be used going forward. Closing this pr.

PR #209, opened 2021-01-25
Covariances

semaphoreP:
Functionality to include covariances in the measurement uncertainties. After the user passes in covariances into the read input, it should automatically do the rest. I checked its functionality by running an OFTI fit with covariances checking it returns roughly the same posterior, but also that the likelihood value for the same set of parameters is essentially the same no matter if we are using sep/pa, or ra/dec with covariances. 

Changes:
 * read_input now can take in `seppa_corr` and `radec_corr` values as input
 * read_input always outputs `quant12_corr` as an output column. The column is all nan if there are no correlated errors.
 * Added test in test_read_input that checks the above output behavior is correct.
 * lnlike.py now can compute a chi2 with a 2x2 covariance matrix if `quant12_corr` is non-zero for a given epoch. 
 * New test module `test_chi2_lnlike.py` where the old `test_chi2lnlike()` test has been moved to, in addition to tests that our formulas for 2x2 chi2 likelihoods are correct by checking against numpy functions. 
 * Function in system.py called `transform_errors` that using Monte Carlo draws to convert errors (and covariances) from one basis to another given a transformation function. 
 * OFTI sampler now handles removing the determinant of the covariance matrix when converting lnprob to chi2.
 * test in `test_OFTI.py` that checks that OFTI works with covariances and that the covariances are handled properly
   * Tests that the likelihood in sep/pa space is the same as when evaluated in ra/dec space when adding in covariances
   * Tests that the posterior is approximately the same between fits in ra/dec space and sep/pa space. 
 * Documentation in the Formatting Input function on how to include covariances.


semaphoreP:
Should look at this after merging in #198 since some of the changes are from that right now.

sblunt:
@semaphoreP looks like there are some merge conflicts, I think primarily with the rvinstruments code I just merged. Would you mind looking through and resolving them? Note that I merged in #198 already.

semaphoreP:
@sblunt I think things are good to go regarding merge conflicts. Maybe @vighnesh-nagpal could check that my changes to read_input.py are good. I'm not sure if there are currently tests/documentation for the RV instruments feature.

sblunt:
Thanks for clearing those up @semaphoreP! Looks good to me.

sblunt (approved):

PR #211, opened 2021-01-26
Add FAQ section with tau docs.

semaphoreP:
New FAQ section of docs, with a new notebook on tau/tp/M. 

One question, I realized tp is named inconsistently. In `orbitize.basis`, it is referred to as t0. Do we have a preferred notation?

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -1,0 +1,21 @@
+.. _faq:
+
+Frequently Asked Questions
+==========================
+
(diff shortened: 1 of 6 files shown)


sblunt:
I believe the tp conflict was resolved when we chatted over zoom, @semaphoreP, but reopen if not. We decided that the preferred notation is Tp.

sblunt (approved):
Looks awesome, thanks for starting this section up @semaphoreP !

PR #216, opened 2021-01-27
Prior KDE Implementation and Tutorial

jorgellop:
I fixed a couple of bugs in priors.py in the KDE implementation, added some features, and added a tutorial notebook


sblunt:
Looks great @jorgellop, thanks so much for working all of this out! I really like the Few small things:
1) please add I/O documentation to the new classes in priors.py (and if not too much trouble, remove commented out debbugging statements)
2) looks like tests are timing out on travis. I'll look into this now as well, but let's make sure tests are passing before we pull in the code.
3) we should think about adding some unit tests. Perhaps have a radvel fit to a simple circular simulated datset, then check that the KDE prior actually produces expected results. @semaphoreP might have some more ideas.

sblunt:
Also need I/O in the new `draw_from_priors()` method

sblunt:
Ok, bugs should be fixed now. Tests are passing on my computer. When you get a chance, add the docstrings and then we'll be good to merge! Thanks again! :)

semaphoreP:
Hey @jorgellop, I cleaned up the code a bit by restoring the tutorial notebooks and update the OFTI code to the newest version (somehow they were unintentionally edited). I think there's a few things left to do:

 * There's a duplicate RV_tutorial.ipynb. Was this duplicated by accident?
 * In your KDE priors tutorial notebook, we can probably clean up the commented out code in the code blocks and the references to your own computer directory. Simplifying the code blocks would make it easier to read
 * At the end, can you link to the MCMC tutorial, and discuss how to link your tutorial to the MCMC tutorial to sample with orbitize
 * Can you add a test to ensure the KDE and nearest ND interpolator work? It can be something like drawing a bunch of samples from a posterior and making sure the drawn samples have a mean/median that's very close to the original distirbution. 
 * Can you add some sort docstrings to the functions in the `NearestNDInterpPrior` and `KDEPrior` classes? And add spaces between functions.

semaphoreP (approved):
I think it should be good now.

Issue #104, opened 2019-05-01
Improvements to running long MCMC chains (restarting and saving)


semaphoreP:
Have an argument that every so many steps, we save the current chains. This would be good for long MCMC runs so we can inspect the chains, or have their progress in case something crashes. 

A minor complication is whether we use the default Results save type, or whether we also need to expose the format of the saved chains in the MCMC API.

semaphoreP:
We should also have it be able to restart from a current position.

semaphoreP:
We should also do this for OFTI. Could be a separate issue.

semaphoreP:
We should append to the HDF5 file rather than writing a completely new file (if possible).

semaphoreP:
We should try to leverage the emcee v3 feature that does this.

semaphoreP:
Note that ptemcee and emcee might need different solutions to this issue.

semaphoreP:
We should make sure to have the ability to run the sampler, and then restart from where we left off if we call `run_sampler` again.

sblunt:
Addressed in PR #217.

PR #217, opened 2021-01-27
Save load

semaphoreP:
Addresses #104. MCMC Sampler can now periodically save new steps to orbitize.results which saves to HDF5 file. Can also create a sampler that resumes from a previous file.


sblunt (approved):
Nice, looks great. Thanks!

Issue #221, opened 2021-03-28
results.Results.load_results backwards compatibility issue with v1.15 - Can't open attribute (can't locate attribute: 'version_number')


wbalmer:
I had saved the posterior running the code below on a previous version of orbitize (I think 1.13?):

hdf5_filename='my_posterior_lab.hdf5'
import os
if os.path.isfile(hdf5_filename):
    os.remove(hdf5_filename)
my_driver.sampler.results.save_results(hdf5_filename)

and tried to load it on the current version 1.15 using:

hdf5_filename='my_posterior_lab.hdf5'
from orbitize import results
loaded_results = results.Results() 
loaded_results.load_results(hdf5_filename)

which previously loaded the results but now gives the error:
`KeyError: "Can't open attribute (can't locate attribute: 'version_number')"`

I uploaded the file I was trying to load: https://github.com/sblunt/orbitize/files/6213994/my_posterior_lab.zip

Is there a way to salvage this posterior I had saved?

sblunt:
Addressed in PR #222.

Issue #219, opened 2021-03-17
Backwards compatability with v1.14 and v.15 features


semaphoreP:
Some new features currently don't work with old versions of orbitize data products (input astrometry/results files). 

  * not having `quant12_corr` or `instrument` in an orbitize data file (the ones with `quant_type` as a field)
  * not having an orbitize version in the results file

We should add some default options to handle missing cases.

wbalmer:
have been loading a hdf5 results file using loaded_results.load_results(hdf5_filename) (file attached) and loaded_results.data is None. Should I revert to 1.14?

[my_posterior_lab.zip](https://github.com/sblunt/orbitize/files/6213994/my_posterior_lab.zip)

semaphoreP:
@wbalmer Hmm.. this seems like a bug. I'm not sure if reverting to a old version will fix it since it's a relatively new feature. Are you trying to access the input astrometry? Probably the suggested workaround is to read in the input astrometry again with `read_input` in your code.

wbalmer:
@semaphoreP Sorry, I realize now I misread and this issue was about the input astrometry, so I opened a separate issue #221 . I'm trying to access the posterior that I had saved previously, the problem seems to be that the results file I had saved doesn't have a version number.

semaphoreP:
@wbalmer Oh no, you were right in interpreting this issue was with the version number! I just misread your comment. I just opened PR #222 to fix the version backwards compatibility issue. Feel free to copy that code or use that branch if you need this to work before we release a new version of orbitize.

semaphoreP:
I'll add that it seems like perhaps the results.data being None is a separate bug. We'll take a closer look!

semaphoreP:
It looks like results.data is being populated for me. @wbalmer, if you still have this problem, can you re-open #223 and point us to a script to reproduce?

PR #222, opened 2021-03-29
error catch for old results without versioning

semaphoreP:
Address issue #221, and partially #219. Now catches error if version number is not in the results files.


sblunt:
Looks good to me, thanks for adding this @semaphoreP.

sblunt (approved):

PR #227, opened 2021-04-02
Updating Driver to accept stellar_mass instead of system_mass for con…

lhirsch238:
…sistency with Sampler.

Modified variable name in driver.py for consistency with sampler.py.


lhirsch238:
Still need to modify the tutorials with the new variable name.

PR #228, opened 2021-04-02
Backwards compatibility for previous orbitize! data formats

semaphoreP:
If users (like me) saved input astrometry with older versions of orbitize, they are missing `quant12_corr` and `instrument` columns. I added a check to populate these columns with default values rather than crashing. Address part of issue #219


vighnesh-nagpal (approved):
Totally didn't think about backwards compatibility when working on the code. Thanks for catching this and making the fix, looks great!

PR #235, opened 2021-06-08
add token for running codeastro tests

semaphoreP:
Run `pytest --mode codeastro` and if all tests pass, you get a secret token at the end. Pretty basic obfuscation but prevents any simple grep of the code to find.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,30 @@
+import pytest
+
+codeastro_mode = False
+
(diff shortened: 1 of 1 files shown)


sblunt:
Yay so fun!! Worked on my machine. I got the token when tests passed, and also broke the tests and confirmed that I did not get the token with broken tests.

sblunt (approved):

PR #236, opened 2021-06-17
fix #234. Also another possible bug.

### LINKED ISSUE #234: results.plot_corner() does not correctly interpret the param_list variable
When using results.plot_corner() and defining a param_list variable to restrict the posteriors displayed, the resulting corner plot is not produced / labelled correctly. It appears to only take note of the first term (? further testing requried) when producing the corner plot. However, it correctly pulls the symbols for each parameter and orders them by the order given. This results in mislabeled plots.

Example:

```
corner_figure_median_95 = m.results.plot_corner(
	param_list = ['sma1', 'ecc1', 'inc1', 'aop1', 'pan1', 'tau1'],
	show_titles = True
)
```

produces
orbits47Uma060421

while

```
corner_figure_median_95 = m.results.plot_corner(
	truths=median_values,
	param_list = ['tau1','sma1','pan1','aop1','ecc1','inc1'],
	show_titles = True
)
```

instead produces
orbits47Uma060321

Note in particular the third variable: in the first image, this is labelled as the position angle of nodes, while in the second it is labelled as the inclination, yet the posterior is the same in both images.

semaphoreP:
Addresses issue #234 (see plots below). Also I noticed that line 319 should be ` if np.std(self.post[:, index_num]) > 0:` because `i` indexes into the `param_list` array, not `self.post`. 

Reproducing the issue #234, now behavior is good:

```
corner_figure_median_95 = result_copy.plot_corner(
	param_list = ['sma1', 'ecc1', 'inc1', 'aop1', 'pan1', 'tau1'],
	show_titles = True
)
```
image

```
corner_figure_median_95 = result_copy.plot_corner(
	param_list = ['tau1','sma1','pan1','aop1','ecc1','inc1'],
	show_titles = True
)
```

image

### DIFF
--- a/orbitize/results.py
+++ b/orbitize/results.py
@@ -316,7 +316,7 @@ class Results(object):
             index_num = np.where(np.array(self.labels) == param)[0][0]
 
             # only plot non-fixed parameters
-            if np.std(self.post[:, i]) > 0:
+            if np.std(self.post[:, index_num]) > 0:
                 param_indices.append(index_num)
                 label_key = param
                 if label_key.startswith('aop') or label_key.startswith('pan') or label_key.startswith('inc'):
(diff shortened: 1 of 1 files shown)


sblunt:
Looks good to me. Once @adj-smith confirms it fixes issue #234, it I'll pull it in.

adj-smith:
Yes, this works for me now as well!

sblunt (approved):

sblunt:
Excellent! Thanks @semaphoreP!

PR #237, opened 2021-06-21
Changebasis

rferrerc:
(Hopefully this is right this time around) Added the conversions.py module for converting between bases and a Jupyter notebook tutorial draft. Also edited sampler.py, results.py and system.py to support the XYZ basis. (Still some corrections need to be made e.g. currently only works for data provided in RA/Dec).


semaphoreP:
This looks fine to me for now, but we should think about how to abstract most of the basis handling code into a separate module rather than having it being scattered throughout system/sampler codes.

PR #239, opened 2021-06-30
fixed a bug that was causing RA/Dec data points to display wrong in...

sblunt:
…orbit plots

@thaynecurrie let me know if this fixes the bug you were running into.

If a user has a data table with RA/Dec data points, the orbit plot wasn't converting and displaying them correctly. Should be fixed now!


semaphoreP (approved):
Looks good to me.

thaynecurrie:
Hi Sarah:

Thanks very much.   Certainly the display differences have been fixed.
I'm testing the results between the radec & seppa input files and ran the
simulations with num_temps=10, num_walkers=10, burn_steps=25 and
total_orbits=10000.   There are some differences in the posterior
distributions depending on whether I use the ra-dec or sep-pa inputs, but I
don't know whether these differences really matter given the sampling I'm
using.

sblunt:
Hey Thayne,

Great, glad the plotting is fixed. My guess is that your chains just aren't fully converged. Here's my usual ptemcee orbitize setup for short arcs:


```
n_temps=20
n_walkers=1000
n_threads=20
total_orbits_MCMC=50000000 # n_walkers x num_steps_per_walker
burn_steps=10000
thin=10
```

PR #240, opened 2021-06-30
fixed chop_chains function to copy original data over when updating r…

TirthDS:
…esults object- originally, data variable gets set to 'None' after chopping resulting in an error when plotting orbits


sblunt:
Thanks for the bugfix, @TirthDS. Can you add a unit test that tests this case? Then I'll pull it in.

sblunt (commented):
Looks great, please add a unit test.

sblunt (changes requested):
Thanks Tirth! To make these tests able to run automatically with Travis CI (or any other application that uses pytest), the functions have to start with `test_`. Ex: `text_verify_results_data()`. Try renaming them and then running `$ py.test` from the command line to make sure all tests are passing.

sblunt (approved):
Awesome, thanks!

Issue #144, opened 2019-10-15
OFTI doesn't handle PAN priors properly


semaphoreP:
While working with someone on an OFTI fit who was trying to limit the range of PAN, they noticed that OFTI still produced samples outside of the prior bounds. This is because OFTI scale and rotate and rotate PAN to be outside of the bounds, but does not check for it.

sblunt:
Good point. I did this intentionally but did not print an error/warning message. I have added the ability to set priors on pan in the RV posts as priors branch, so I will fix this at latest when I pull that code in.

sblunt:
Partially addressed in PR #186. Still need to implement the ability to apply arbitrary PAN priors in OFTI

sblunt:
Finally fixed in PR #245

PR #245, opened 2021-07-21
account for user-defined priors on PAN

sblunt:
Hey all, I've finally fixed issue #144! Check it out and let me know what you think. Functionality and a few new unit tests.

(Originally did this on a different branch, but built it off of develop instead of main by accident. Then decided it would be simpler to just make a new branch rather than rebasing the old one since it was so little code)


semaphoreP (commented):
Nice looks good. Did you also mean to delete the requirements badge?

sblunt:
Yeah, I don't like that badge XD

PR #246, opened 2021-07-21
Fixed dates in HD4747 data file.

lhirsch238:
The dates were indeed wrong in the HD4747.csv file. I re-downloaded the data from the Crepp et al. 2016 paper and used astropy.time.Time to convert jd to mjd, paying attention to the formats used in the paper tables (for RV data, JD-2440000 and for astrometry data, JD-2450000).


sblunt (approved):
Thanks Lea!

PR #249, opened 2021-08-06
Multiplebasis

TirthDS:
Contains API for multiple basis sets (assigning priors and making conversions).

### DIFF
--- a/orbitize/basis.py
+++ b/orbitize/basis.py
@@ -1,23 +1,761 @@
 import numpy as np
 import astropy.units as u
-
+import astropy.constants as consts
 import warnings # remove when functions are depreciated
+import abc
 
-def tau_to_t0(tau, ref_epoch, period, after_date=None):
+from orbitize import priors, hipparcos
+from scipy.optimize import fsolve
+
+class Basis(abc.ABC):
+    '''
+    Abstract base class for different basis sets. All new basis objects should inherit from
+    this class. This class is meant to control how priors are assigned to various basis sets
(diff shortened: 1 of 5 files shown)


sblunt:
Looks really good so far. I really like the simplicity of the Basis classes. I think we should discuss how we want to move ahead with treating nuisance parameters like the RV instrumental params, Hip IAD params, etc. I like the way you've done it now. Are there other nuisance params we expect to need to take into account in the future? Will this way of passing them in as kwargs to the Basis object become untenable at some point? Not sure, just starting the conversation!

semaphoreP:
Yeah looks great! Nice work abstracting away all of the basis stuff into its own class. I think regarding instruments, in the future, we might want to include uncertainty in direct imaging instrument calibrations (so not just RV instruments). But I think it's ok to add additional nuisance parameters in the future as optional arguments, as long as we program it to assume that there may be additional optional parameters in the future.

sblunt:
Cool, sounds good.

TirthDS:
As a note, XYZ conversions on a very limited set of parameters don't seem to work even after filtering out orbits with ecc >= 1. There is always about 1-2 orbits whose parameters do not match up with the conversions.

semaphoreP:
Hey @TirthDS, I see a rouge `import pdb` in the code, but otherwise, it's looking great. Are we ready to merge soon, or should we discuss more?

TirthDS:
Hi, we should be ready to merge soon, but I am still a little concerned about the correctness of the XYZ conversions after running the tests and was hoping to discuss with @rferrerc about it.

PR #255, opened 2021-08-19
Test function for check_prior_support() function

adj-smith:
See pull request #253 for the function being tested by this script.

### DIFF
--- a/tests/test_checkpriorsupport.py
+++ b/tests/test_checkpriorsupport.py
@@ -1,0 +1,58 @@
+import numpy as np
+import orbitize
+import orbitize.sampler as sampler
+from orbitize import driver
+import random
+
+
+def test_check_prior_support(PriorChanges=False):
+	'''
+	Test the check_prior_support() function to ensure it behaves correctly.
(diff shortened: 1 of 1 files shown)


semaphoreP:
hey @adj-smith, nice test. Just a few things:

* Can you commit this to your `patch-1` branch, so that PR #253 automatically gets updated. That way we don't have two separate PRs (and we'll just close this one)
* For automated tests, we don't expect a user to look at the outputs unless things fail. So instead, we use assert statements or exceptions for pass/fail: no assertion fails or exceptions on test pass, and either an assertion fail or an uncaught exception for test tail. Can you modify your test to do that? 
* We typically wrap the function calls with `if __name__ == "__main__" at the bottom of the file. This prevents the test being called twice in the pytest framework, and the bottom call only runs if a user runs it manually from the command line. 

Check out some of our other tests for examples. Thanks!

adj-smith:
Done. I was trying to work out how to include the test in my other pull request. Now I know. Thanks!

semaphoreP:
thanks! closing this PR to review changes in #253 instead.

PR #258, opened 2021-08-25
Hipparcos api

sblunt:
Not completely ready yet, but here are the changes I made so far:

1. Throw error when Sn != 5
2. Allow arbitrary alpha0, dec0 epoch
3. Pass Hipparcos object to `System `directly.
4. `test_abs_astrometry()` now works to machine precision

Still to do:
1. add code to throw out rejected scan measurements
2. Arrays -> dictionaries for holding parameters (not sure about this but including it here so I think about it)
3. Add plotting capability
4. include test to recover fitted astrometric params (I have this just have to reformat it a bit)


semaphoreP (approved):
Looks good! My only nitpick would be to raise a ValueError rather than a generic Exception for the 5-parameter fit check (and corresponding check for a ValueError exception rather than any generic exception in the test).

sblunt:
Good call; fixed! Thanks!

PR #260, opened 2021-08-26
RV Plotting Fix

TirthDS:
1. Fixes the way rv instruments are chosen in the `plot_orbits()` function (so that it doesn't ignore the default rv instrument 'defrv').
2. Can plot just a single orbit
3. Can plot astrometry instruments
4. When plotting rv, we take the 'max' of the likelihoods instead since they are all negative


semaphoreP (approved):
Looks good! Thanks for making those changes!

semaphoreP:
Tests passed for me on this branch as well

Issue #256, opened 2021-08-24
`results.load_results()` incorrectly parsing data from HDF5


adj-smith:
Possibly related to issue #252...?

After a recent update, when loading a saved MCMC posterior with `results.load_results`, the resulting `results` object does not appear to be generated correctly. In particular, the `results.data` attribute is _not_ a proper orbitize! `data_table`. To make matters worse, the strings in the data are not read as standard Python strings, but instead become bytewise literal strings. This causes trouble when trying to later use functions such as `plot_orbits()` as the data types are incorrect. This causes such lovely errors as

```
--> 646                 plt.scatter(Time(astr_epochs,format='mjd').decimalyear,sep_data,s=10,marker='*',c='purple',zorder=10)
ValueError: x and y must be the same size
```

in `plot_orbits()`, which I was able to chase down to various `if` statements trying to reconcile whether `'radec' == b'radec'` (circa line 530), which is naturally `False` and causes problems downstream.

I ended up doing a workaround and simply reassigning my `results.data` from a `data_table` parsed directly from my stored data, which corrected the issue. Naturally this doesn't work in every case however.

TirthDS:
One suggestion I have is before storing the `results.data` attribute, change the dtype of all the columns containing objects with type 'SN' to

wbalmer:
was also encountering this issue - the fix worked for me too :)

sblunt:
Addressed in PR #262

PR #262, opened 2021-08-27
use astropy tables in results to address #256

semaphoreP:
The text fields in results.data should now be str and not bytes. Addresses #256, and possibly also @252.


sblunt (approved):
Great, thanks for tracking this down @semaphoreP.

PR #266, opened 2021-09-01
Indices to param idx

sblunt:
Decent-sized API change. Rather than relying on a constant ordering of parameters under the hood, _orbitize!_ now uses the `param_idx` attribute, a dictionary matching parameter labels to their indices in the array of fitting parameter values.


semaphoreP (approved):
Remove binary file and run tests and we are good.

Issue #158, opened 2020-02-08
Add omega/Omega documentation


sblunt:
Brendan Bowler suggested we add some more explicit documentation about the relationship between tau/orbital period/EPP and the geometric interpretation of omega/Omega. I’d also like to add something explaining the geometric inclination prior since that’s a question I get asked a lot.

semaphoreP:
related to issue #106 also

sblunt:
Add to FAQ section.

sblunt:
Addressed in PR #267

PR #267, opened 2021-09-02
add faq on inc, aop, pan

semaphoreP:
Mostly refer reader to show me the orbit, but also new page on how inc/pan define the orbital plane, and why sine priors on inc.


semaphoreP:
Hopefully addresses issue #158

sblunt:
Nice, this is awesome. I made two small changes-- 1) removed an unused matplotlib import; 2)slightly changed your explanation of Omega to one that makes more intuitive sense to me. But my explanation is a bit wordy. Let me know what you think; happy to change it back.

semaphoreP:
Looks good to me!

PR #268, opened 2021-09-02
add full hipparcos tutorial

sblunt:


semaphoreP (approved):
Maybe add a sentence that Gaia astrometry is needed to recreate the mass estimates from Nielsen et al. Otherwise works for me after I deleted a rouge plistlib import.

Issue #248, opened 2021-08-03
Bug in setting of track_planet_perturbations


sblunt:
In this line:

```
        self.track_planet_perturbs = self.fit_secondary_mass and \
                                     ((len(self.radec[1]) + len(self.seppa[1]) + len(self.rv[1]) < len(data_table)) or \
                                      (self.num_secondary_bodies > 1))
```

I believe  ```len(self.rv[1])``` should be ```len(self.rv[0])```. Don't think this is messing anyone up right now but we should fix this.

sblunt:
addressed in PR #274 (and see the conversation in that issue)

PR #274, opened 2021-09-04
address #248

sblunt:
We should set track_planet_perturbs to True if the user wants to fit for a dynamical mass AND

1. there is data other than relative astrometry and ** stellar ** RVs (i.e. absolute astrometry of the primary)
2. there is more than one secondary body


sblunt:
Just kidding. Here's the (hopefully) correct logic:

We should set track_planet_perturbs to True if the user wants to fit for a dynamical mass AND

- there is stellar absolute astrometry data OR
- there is more than one secondary body

semaphoreP (approved):
looks good

Issue #281, opened 2021-09-18
Ensure backwards compatibility with v1 orbitize results. 


semaphoreP:
Use default values for fields that don't exist in v1.

semaphoreP:
Addressed in PR #284. Re-open if there continues to be issues.

PR #284, opened 2021-09-27
V1 results load compatibility

sblunt:
Fixed loading of old results & added a unit test so we don't keep breaking this in the future


sblunt:
Addresses issue #281

semaphoreP (approved):
Looks great!

PR #287, opened 2021-09-29
Plotting api

sblunt:
Some reasonably big changes to the Results class, including making System an attribute of Results and separating out plotting routines into a separate module. I also reviewed the documentation for all of the modules and made some minor changes


orbitize/sampler.py
```diff
@@ -34,15 +36,17 @@ def __init__(self, system, like='chi2_lnlike', custom_lnlike=None):
             self.lnlike = getattr(orbitize.lnlike, like)
 
         self.custom_lnlike = custom_lnlike
+        self.use_c = use_c
```
semaphoreP, line 39: Not related to this PR, but what is the purpose of having `self.use_c` and `use_c` in the arguments of `run_sampler` and `_logl`. Some of the behavior has it as default `None` and some as default `True`. Is this an opportunity to override default values, and if so, do we actually implement it like that? Do we need all of them?

orbitize/system.py
```diff
@@ -269,6 +313,11 @@ def compute_all_orbits(self, params_arr, epochs=None, comp_rebound=False):
                     radial velocities at each epoch.
 
         """
+
+        if use_c == None:
```
semaphoreP, line 317: Similar question here with the None vs default values for `use_c` and `use_gpu`. Perhaps we could just set one global variable in orbitize for these settings?

sblunt: Yeah, this is a good point. This code has already been merged into main like you said, so I say we open an issue and fix it once the open PRs are good to go.

Issue #208, opened 2021-01-25
Docs page for new collaborators


sblunt:
We should add a docs page that summarizes best practices for contributing to orbitize!, starting with our collaborator_guidelines.md file.

Issue #164, opened 2020-04-09
Expand documentation on fixing priors 


semaphoreP:
We should put an explicit example on how to fix a prior to a value (currently just mentions that parallax is fixed, but an explicit example would be nice).

sblunt:
Addressed in PR #294

PR #294, opened 2021-10-06
Docs facelift

sblunt:
- cleaned up some html display stuff
- added new modules to api documentation toctree
- added page for new collaborators (issue #208)
- added detail to modifying priors tutorial about fixing parameters (issue #164)


docs/tutorials.rst
```diff
@@ -40,12 +43,21 @@ us if you are still confused).
...
    tutorials/Modifying_Priors.ipynb
-   tutorials/Using_nonOrbitize_Posteriors_as_Priors.ipynb
    tutorials/Plotting_tutorial.ipynb
    tutorials/MCMC_vs_OFTI.ipynb
    tutorials/Modifying_MCMC_initial_positions.ipynb
 
+**Feature Tutorials:**   
+
+.. toctree::
+   :maxdepth: 1
```
semaphoreP, line 54: I think you can add labels to the toctrees btw if it helps with separating them: `    :caption: Setup`

semaphoreP (approved):
Looks good. I just added one comment regarding a way to label toctrees if it is useful.

docs/tutorials.rst
sblunt: Thanks! The reason I separated the tutorials into three subsets is because we now have categories of "tutorials that teach you to use the code in the default use case," "tutorials that teach you more about the underlying API," and "tutorials that teach you to use specific new features." Do you think those descriptors would be helpful to add as captions?

PR #299, opened 2021-10-12
Amend windows install instructions

sblunt:


semaphoreP (approved):
Looks good! We should consider having a --skip-rebound or other similar features in the setup in the future if this becomes a hassle.

Issue #57, opened 2018-09-05
Speedups for OFTI algorithm


sblunt:
This might be something @logan-pearce is interested in working on. In particular, I think we should implement the following:
- use a scipy optimization routine to estimate the minimum chi square for a given dataset, then use that to calculate a modified chi square for each tested orbit.
- adaptively modify priors as regions of parameter space are ruled out.

sblunt:
Sounds fun! Lol. No worries. This isn't a high priority item, so one or both of these speedups may still need to be implemented next semester if you have more time then. I'll put them on my todo list and keep you updated. 
Good luck with applications, classes, and papers! :) You got this!!!

sblunt:
Update: 
- we've decided that the second bullet point in initial issue should be discussed in documentation, but not explicitly worked into the code.
- look into stochastic gradient descent for bullet point 1

sblunt:
@lhirsch238 potentially interested in this as a student project

semaphoreP:
Also to inform MCMC chain starting locations to improve convergence.

sblunt:
Added to MCMC starting positions tutorial in PR #301. Items still on the todo list for this issue include using the minimizer results to speed up OFTI, as well as exploring which minimzers work best for orbits.

PR #301, opened 2021-10-25
Adding scipy.optimize.minimize to modify initial positions

Mireya-A:
I added code that allowed users to use scipy.optimize.minimize to the Modifying Initial Parameters Tutorial as a way to create initial positions for MCMC with a starting array of guess parameters for their orbit.


sblunt (approved):
Looks great Mireya, thanks for adding this in!

sblunt:
Partially addresses #57

PR #312, opened 2022-01-28
Chisquared log

Mireya-A:
Modified the lnlike.py to include a log version of chi-squared. This implementation defines separation chi-squared in log scale, and defines position angle chi-sq using complex phase representation.


sblunt (commented):
Hey Mireya,

This looks great! Thanks for coding it up! 

A few minor comments:
1. First, remove the .DS_Store file.
2. Could you add a unit test that uses the new functionary? It would be cool in particular to compare the distribution of orbits accepted for an OFTI run using the standard chi2 vs the log chi2.

sblunt (approved):

Issue #324, opened 2022-08-19
Error with reading IAD data in load_results


tomasstolker:
Hi @sblunt!

I encountered a small issue when processing results that were obtained with IAD data, which seems to be related to how the IAD data are stored in the HDF5 file and then again extracted by `Results`. It might be because I ran the MCMC on a Linux machine and then processed the results on Mac.

When the data are read from the HDF5 file and temporarily stored in the `thisisprettyhackysorrylmao` (😄) text file, there are many `<NULL>` added to the actual data, which then causes an error when that temporary file is read by `np.loadtxt` in `HipparcosLogProb`.

In the `_save` method of `HipparcosLogProb` the full IAD data file is stored as string dataset. A solution might be to store the actual data as floats. In case any details from the IAD header should also be stored then these could perhaps be included as attributes of the HDF5 dataset.

Thanks for all your work on this great package!

sblunt:
Hey @tomasstolker, thanks for the bug report (and for very kindly pointing out my silly naming scheme). I can reproduce this error by running an MCMC on a linux machine then trying to load it on my mac. Looks like there is some additional type conversion that needs to happen to make sure all the OS play nice together. I'm on it!

sblunt:
Fix in PR #325. @tomasstolker, please let me know if this fixes the issue for you. It did for my setup.

sblunt:
Fixed in #324.

PR #325, opened 2022-08-30
fix for issue #324

sblunt:


semaphoreP (approved):
Ah, good ol' string formats. As long as it doesn't break existing tests, I'm satisfied.

sblunt:
Oh right, tests! XD I just added another unit test so we're now testing saving/loading for both formats of IAD files we accept, which broke, so I fixed that too. I checked that .hdf5 files generated by both of those tests on a linux machine can be loaded correctly on my mac.

semaphoreP:
Great! Looks good then. Feel free to merge

sblunt:
Cool. Will do once @sefffal has a chance to test.

sefffal:
All good!

tomasstolker:
Thanks very much! Works for me as well!

PR #327, opened 2022-12-08
Dynesty

thea-mckenna:


sblunt:
Awesome work Thea!!!

Dynesty Branch Comments:
- nested_test -> move to test/end-to-end-tests
- prior_test -> remove from orbitize directory (but save somewhere on your computer)
- dynesty_tutorial -> move to docs/tutorials
- In priors.py:
1. Remove commented out code
2. Leave a space between the # and the start of a comment
3. In transform_samples documentation:
4. Mention the bounds of u
5. Add Args and Returns to docstring
6. Also, we need to implement transform_samples for the other prior classes. For now, let’s just add a NotImplementedError
- Add a unit test that makes sure NestedSampler API actually runs. Maybe a good way to do this would be to fix all parameters except, say, eccentricity, then let that run to completion. Let’s discuss. This can be the last thing we add.
- In sampler.py:
1. Use either ‘’’ or “”” for block comments (I prefer “””)
2. Add “Thea McKenna, 2022”
3. In ptform, write out bounds for u, and add Args and Returns statements
4. Stick to 80 characters per line
5. wt_kwargsdict is unused (should be passed into sampler)
6. total_orbits is unused (can be removed)

sblunt:
Code is good to go from my perspective once the tutorial is finished!

sblunt:
Ok, assuming tests pass (which they do on my machine), I think this branch is good to go!

sblunt:
The GJ 504 end-to-end test that we currently have is taking absolutely forever, but the tests in the tutorial run relatively quickly. We should add a different end-to-end test that we can compare more easily against MCMC.

semaphoreP (approved):
Looks good to me at a cursory glance through the docs and tests

Issue #330, opened 2023-01-23
Float function of NumPy does not exist?


tomasstolker:
This line:
https://github.com/sblunt/orbitize/blob/910d4065e4ff702d076426c9ddf64b3504631f6a/orbitize/read_input.py#L237

Causes the following error with my installation:
`AttributeError: module 'numpy' has no attribute 'float'`

I think because `numpy.float` does not seem to exist:
https://numpy.org/doc/stable/user/basics.types.html

Perhaps using `numpy.float_` solves it? Or maybe using `numpy.array` and then check the `dtype` afterwards (instead of using the `try`-`except`).

tomasstolker:
I think that this issue is about the same error actually: #329

tomasstolker:
The problem is related to `numpy` version 1.24 (`np.float` was deprecated). Downgrading to version 1.23 solved the issue!

sblunt:
Addressed in PR #331. Thanks @tomasstolker!

Issue #329, opened 2023-01-20
Problem reading epoch in the input file. Epoch should be given in MJD


AlessioPop:
**reading epoch in the input file raises errors**

I used the same Data provided by the documentation. If I mark the provided Data (MJD) with # in the CSV file I get an empty Table (see Image below).

 
![helpYOU](https://user-images.githubusercontent.com/120112894/213755632-d7de5e04-52df-49cc-b135-a70bc527e676.PNG)

Meaning that the path to the file is working properly. However, when using the provided Data (MJD) I get the following error: 

![helpihelpi](https://user-images.githubusercontent.com/120112894/213754729-9b37760b-bd9e-46b4-a1aa-ef90817edf17.PNG)

**My System:**
 - OS: Linux
 - orbitize! Version 2.1.2

AlessioPop:
orbitize does not support the latest version of numpy, so one should downgrade to an older version of numpy to solve the issue.

sblunt:
Thanks for the bug report!

Addressed in PR #331. Thanks @tomasstolker!

PR #331, opened 2023-01-24
Updated deprecated numpy functions

tomasstolker:
This is a minor PR that fixes the issue related to several deprecated `numpy` functions (in version 1.24) for certain data types (see issues #330 and #329).

I have updated the use of `np.float`, `np.int`, and `np.str`. This fix should also work with `numpy` versions older than 1.24.

Not related, but I also added a warning in `plot_orbits` related to the `rv_time_series` parameter, just for convenience.

Let me know in case you have any feedback 😊


sblunt:
Thanks so much for the fix, @tomasstolker, and sorry it took me a while to get around to looking at this. I was able to reproduce the error referenced in #329 and #330, and this fixed it. Thanks for the warning addition as well.

sblunt (approved):

PR #332, opened 2023-02-02
bugfix for making orbit plots when we have a pl rv

sblunt:


semaphoreP:
Can you describe the bug and fix? I'm not sure what to be looking for.

sblunt:
@semaphoreP the plotting code was failing when there is a secondary RV measurement in the data file. This fixes that failure.

semaphoreP (approved):

Issue #335, opened 2023-03-22
Type issue with num_period


tomasstolker:
Hey @sblunt, just a minor issue I ran into: when setting `after_date` in `tau_to_tp`, it causes an error if the input are arrays instead of floats because `num_periods = int()` requires a float. This would fix it I think:

```
if isinstance(num_periods, np.ndarray):
  num_periods = np.ceil(num_periods).astype('int')
else:
   num_periods = int(np.ceil(num_periods))
```

semaphoreP:
Ah, originally, the function definition does not allow arrays for `after_date` but this can be easily changed. PR #336 adds this functionality (we don't even need the `else` because the np.float64 output by np.ceil for single values has the `astype` function as well.

tomasstolker:
Great, thanks a lot!

Sorry I didn't explain it properly I realize now. The problem that I ran into was a case where `after_date` is a float instead of None, while `tau`, `ref_epoch`, and `period` are arrays. In that case `num_periods` became an array and caused the error.

I agree that for `after_date` it is typically not required to use an array but your fix seems fine and adds even a bit more flexibility to the function 👍.

Thanks again 😊

sblunt:
fixed in #336

PR #336, opened 2023-03-22
after_date can accept multiple values now

semaphoreP:
Addressing #335, the `after_date` keyword in `basis.tau_to_tp` can now accept numpy arrays. Test included to check it works.


sblunt (approved):
Looks great, thanks for fixing this Jason! Just pulled in latest changes from main, and will pull in once CI tests run

PR #338, opened 2023-06-17
fix orbitize tests

semaphoreP:
The truth values have an extra 1/sqrt(2pi) that we aren't accounting for. I multiplied it to the truth values and the test should pass now.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ astropy>=4
 numpy
 scipy
 emcee>=3
-ptemcee
+ptemcee @ git+https://github.com/sblunt/ptemcee@main
 matplotlib
 corner
 h5py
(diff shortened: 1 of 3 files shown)


semaphoreP:
I also fixed the REBOUND test that is crashing for code/astro participants

sblunt:
Thanks very much Jason! I'm still setting up the new CI system so CI tests aren't passing on this PR, but pytests just went through on my machine with python 3.10. I forked a copy of `ptemcee` in order to make it compatible with python 3.10, which as we discussed offline we are seeing as a medium-term solution to `ptemcee` no longer being maintained. Long-term solution is more and better backend algorithms!

sblunt (approved):

PR #339, opened 2023-06-21
GitHub actions

sblunt:
- Removed radvel as a dependency, and separated the radvel/orbitize separation compute function into its own repo. 
- Got gh actions CI set up, and tests passing
- did a little pep8 cleanup for funsies


sblunt:
I also turned on readthedocs building for pull requests, and set it to error if any notebook cells fail, so our new workflow will run through all of the notebooks when PRing to main.

semaphoreP:
Looks good. please add to the release notes this is a breaking changing in deprecating radvel utils.

Issue #303, opened 2021-11-02
OFTI should have a time warning


sblunt:
If OFTI goes more than x mins without accepting any orbits, new users often think the software isn't working. We should return some sort of warning, like "OFTI is taking a while, you may want to consider MCMC, check out the MCMC vs OFTI tutorial"

sblunt:
Addressed in #355, released in v3.

PR #345, opened 2023-09-24
OFTI time warning 

michaelkmpoon:
If OFTI doesn't accept its first orbit in 60 seconds, it will now print a warning statement following issue #303.


sblunt:
This PR transferred to #355 (using this code, plus minor changes)

Issue #79, opened 2018-10-30
MCMC return autocorrelation time


semaphoreP:
Related to issue #37. We should have some approved tools for assessing convergence of MCMC walkers, since it is probably the trickiest part of MCMC orbit fitting. Similarly, we should expand the MCMC tutorial. 

- [ ] function that returns the autocorrelation time of the chains
- [ ] documentation in the MCMC tutorial on how to use the above feature to assess convergence

zhexingli:
Is mcmc convergence criteria now incorporated into orbitize? Just being curious

semaphoreP:
Not yet, but check out section 2.3.2 of the orbitize! paper for a summary of how we assess convergence: https://arxiv.org/abs/1910.01756

sblunt:
More discussion added to MCMC tutorial in PR #79

sblunt:
Further discussion added in PR #346. Closing this now; feel free to reopen if more functionality is desired.

PR #346, opened 2023-09-24
Add autocorrelation time estimate

michaelkmpoon:
Related to issue #79.

Added an autocorrelation time estimate (using emcee.autocorr.integrated_time) to the MCMC_tutorial docs, as another way to assess convergence.


sblunt:
Thanks for adding this @michaelkmpoon!!

Issue #291, opened 2021-10-05
Make plots to display Hipparcos/Gaia fit results


sblunt:
Two that I'd like to see:

(1) orbit tracks of absolute ra, dec position of star minus hip0 position (keyword to incl plx or not)
(2) orbit tracks of pmra, pmdec of star

semaphoreP:
Could be 1D histograms showing predicted and measured values? Maybe better to show PMa, rather than absolute astrometry.

sblunt:
Lea has some code, check with her when I get to this. Plot ideas: Fig 1 of https://arxiv.org/pdf/1808.06257.pdf

wbalmer:
I'd love to be able to replicate the third panel of figure 2 here (https://arxiv.org/pdf/1912.02565.pdf) i.e. delta mu_RA and Dec over time, plotting the hip and gaia measurements (almost like the RV plot?)

wbalmer:
![orbitize_hgca_plot](https://github.com/sblunt/orbitize/assets/39763022/a24cbfbe-8ef7-4fec-978d-e5acae86d6da)

I have a start on this (finally). I'll try to fold it into a nice function. Should it be separate from the plot_orbits function, or a yes/no option within it?

sblunt:
Awesome! This looks great. Thanks William! I personally feel like our orbit plots are getting a little bloated, so we should keep this separate.

PR #348, opened 2023-10-02
Proper motion vs epoch plotting function

wbalmer:
Here's a function to plot the proper motion vs time, with curves color-coded by a cbar param. Still working on a stellar absolute position plot, but this should address the 2nd point Sarah raised in #291 . Also included an example in the HGCA tutorial notebook.


semaphoreP:
This looks good. I think it is worth mentioning somewhere that the trajectories aren't *exactly* what we are fitting in the likelihood, since we actually try to forward model Hip/Gaia per epoch and infer the differential proper motion. But they are close enough for visualization purposes. This is why the Hinkley+ 2023 paper only shows histograms.

wbalmer:
Right. I'll add some disclaimer text to the be printed during the function call and included in the docstring.

semaphoreP:
I would say putting it in the docs/tutorial would be even better. Just thinking out loud, I think a succinct way of describing it is that we're actually fitting the time-averaged proper motions and not the instantaneous ones plotted. Might be more clear than waving hands about forward modeling the per-epoch scans.

wbalmer:
Gotcha! Let me know if that message in the function and tutorial notebook is any better.

semaphoreP:
Looks good! Merged now.

PR #352, opened 2024-01-06
update hgca edr3 to stable vizier url, safer fits.open

wbalmer:
Hi Jason! These might be unnecessary, but I was setting up a few orbit runs on the cluster here at jhu and I ended up running into some file not found and an I/O error with fits. I updated where the hgca was being downloaded from to the vizier FTP source and added the "ignore_missing_end" kwarg to the fits.open statement to make this download and read step more stable.


semaphoreP:
Sounds good. Can you merge into the regular `hgca` branch of orbitize instead? We're not using this branch actively anymore. Going to close this PR and wait for the new one.

Issue #357, opened 2024-02-05
Fitting with two planets using rebound throws a fatal error


bailer-jones:
**Summary**
I think a 2D array needs to be recast as a 3D one when using rebound for the forward model.

**The issue**
I am using orbitize (version 2.2.2 in python 3.9.18 on MacOS 13.3.1) to fit a system of two planets with RA,Dec data using the rebound feature. When I do this the sampler (using MCMC) gives the following error:

     File "/opt/anaconda3/envs/rebound/lib/python3.9/site-packages/orbitize/system.py", line 540, in compute_model
     model[self.radec[body_num], 0] = raoff[self.radec[body_num], body_num, :]  # N_epochs x N_bodies x N_orbits
     IndexError: too many indices for array: array is 2-dimensional, but 3 were indexed

**To Reproduce**
sys = system.System(num_secondary_bodies=2, 
                    use_rebound=True, 
                    fit_secondary_mass=True,
                    data_table=dat, 
                    tau_ref_epoch=timeref,
                    stellar_or_system_mass=mass_star/msun, mass_err=0,
                    plx=1.0, plx_err=0.0)
lab = sys.param_idx
mcmcpar = {
    'num_temps': 1,
    'num_walkers': 20,
    'num_threads': 1,
    'num_steps': 1000,
    'num_burnin': 0,
    'num_thin': 1
}
mySampler = sampler.MCMC(sys, num_temps=mcmcpar['num_temps'], num_walkers=mcmcpar['num_walkers'])
blah = mySampler.run_sampler(total_orbits=mcmcpar['num_steps']*mcmcpar['num_walkers'], 
                             burn_steps=mcmcpar['num_burnin'], thin=mcmcpar['num_thin'])

In my case dat comprises 50 simulated astrometric measurements on just one planet, and various priors are set. but I don't think the exact details here are relevant.

**Apparent cause**
I think the issue is in the system.py file in the compute_all_orbits() function.
(Note that when this is called by the sampler it reports having n_orbits=1 inside.) Line 350, which is only called when we are opting to use rebound, is

    raoff, deoff, vz = nbody.calc_orbit(epochs, sma, ecc, inc, argp, lan, tau, plx, mtot, tau_ref_epoch=self.tau_ref_epoch, m_pl=m_pl, output_star=True)

This returns raoff, deoff, (and vz) as 2D arrays of size N_epochs x N_bodies. These are passed back to line 505 of compute_model():

[...]

PR #360, opened 2024-02-12
fix nbody indexing error & add unit test

sblunt:
Fix for issue #357


sblunt:
The nbody code was throwing an indexing error when used in an MCMC run (issue #357). I fixed the error and added a unit test. I also made the choice of nbody backend integrator a keyword.

orbitize/nbody.py
```diff
@@ -16,104 +28,115 @@ def calc_orbit(
         inc (np.array): inclination [radians]
         aop (np.array): argument of periastron [radians]
         pan (np.array): longitude of the ascending node [radians]
-        tau (np.array): epoch of periastron passage in fraction of orbital period 
+        tau (np.array): epoch of periastron passage in fraction of orbital period
             past MJD=0 [0,1]
         plx (np.array): parallax [mas]
-        mtot (np.array): total mass of the two-body orbit (M_* + M_planet) 
+        mtot (np.array): total mass of the two-body orbit (M_* + M_planet)
```
semaphoreP, line 34: should `mtot` be an float, instead of an array?

orbitize/nbody.py
```diff
@@ -16,104 +28,115 @@ def calc_orbit(
...
         m_pl (np.array, optional): mass of the planets[units]
         output_star (bool): if True, also return the position of the star
             relative to the barycenter.
+        integrator (str): value to set for rebound.sim.integrator. Default "ias15"
 
     Returns:
         3-tuple:
 
-            raoff (np.array): array-like (n_dates x n_orbs) of RA offsets between 
+            raoff (np.array): array-like (n_dates x n_bodies x n_orbs) of RA offsets between
```
semaphoreP, line 46: is n_orbs always 1? if so, we should specific that.

semaphoreP (approved):
Unit test looks good. I just have a couple questions on the docstring

orbitize/nbody.py
sblunt: yes, good catch

orbitize/nbody.py
sblunt: yes, good catch. Thanks!

sblunt:
Thanks for catching the docstring errors. I updated it to be consistent with the way the function works.

PR #361, opened 2024-02-29
orbitize plot for companion RVs and plot more host star RV sols

chihchunhsu:
orbitize plot for companion RVs, and add more orbits to plot host star RV solutions


sblunt:
Thanks so much for PRing these changes Dino! This PR has now been migrated to #366.

PR #362, opened 2024-02-29
Orbitize manual

sofiacovarrubias:
Adding a manual section to the orbitize! docs


sblunt:
This looks great so far Sofia! Thanks for doing this. I'll merge this branch in now, but also wanted to note as a todo item for later that it would be great to add in the omega/Omega visualization plot I had in my thesis chapter if possible. I may be biased but I like that plot a lot XD

sblunt (approved):

Issue #347, opened 2023-10-02
Standard to XYZ documentation needs a minor fix


michaelkmpoon:
**Describe the bug**
I think orbitize.basis.XYZ.standard_to_xyz has some incorrect documentation. To convert from standard orbital elements to XYZ for more than 1 set of parameters, the documentation says that the first dimension must be the number of orbital parameter sets. 

![image](https://github.com/sblunt/orbitize/assets/35639046/912e8916-9f4e-4d9b-a09d-40117ebf78a7)

**To Reproduce**
Here is a simple example that shows the error, as well as the how instead I got it to work (by transposing the array). Instead, it should be the second dimension that is the number of orbital parameter sets.

![image](https://github.com/sblunt/orbitize/assets/35639046/3de0a29c-f8e5-4e5f-9804-08d752577163)

This is with orbitize! Version 2.2.2.

sblunt:
Addressed in #363.

Issue #358, opened 2024-02-05
Unclear which mass to specify for fitting two planets with astrometry


bailer-jones:
It is not clear to me from the documentation or code which mass I should specify (stellar or system) for the configuration specified in issue 357, namely

    sys = system.System(num_secondary_bodies=2, 
                    use_rebound=True, 
                    fit_secondary_mass=True,
                    data_table=dat, 
                    tau_ref_epoch=timeref,
                    stellar_or_system_mass=mass_star/msun, mass_err=0, 
                    plx=1.0, plx_err=0.0) 

In
https://orbitize.readthedocs.io/en/latest/system.html
it states: 
"stellar_or_system_mass (float) – mass of the primary star (if fitting
for dynamical masses of both components) or total system mass (if
fitting using relative astrometry only) [M_sol]"
However, this is not unambiguous to me, because "fitting for both components" and 
"using relative astrometry" are not mutually exclusive.
When solving for one planet using relative astrometry, then I understand we give the total system mass (which makes sense).
But if we have relative astrometry on one or more planets and are fitting their masses, which mass do we specify in system.System() above? (I my particular case I need to "fit" for the masses, even though I may keep them constant, as I need to include the perturbing effect of one on the other. I also only provide data for one planet, but that should not be relevant.)

sblunt:
Good point, and thanks for raising this issue! In general, if you're fitting for the dynamical masses of the planets, `stellar_or_system_mass` refers to the mass of the primary star (not the total mass). I'll amend the documentation in the docstring you referenced.

bailer-jones:
Thanks for the clarification :-)

sblunt:
Addressed in #363

Issue #351, opened 2023-12-15
Raised Error when using RV data


juague135:
Orbitize raises an exception when attempting to use a dataset with RV data. "Exception:  If including RV data in orbit fit, set fit_secondary_mass=True"

semaphoreP:
Note that this is the case of only planetary RV

sblunt:
Hey Juan, finally getting around to addressing this bug! I think in the meantime, I must have fixed this somewhere along the line, because I can't reproduce it any more. I was fiddling with the companion rv implementation a lot around this time, so I'm sure it's buried somewhere in all the commit messages. I added a unit test to check this scenario and make sure it stays fixed in the future (in #363). Please reopen if you run into this issue again!

PR #363, opened 2024-02-29
Fix as many small bugs as i can

sblunt:
Fixing as many "easy" bugs as I can in the next 24 hours challenge!

So far, the list includes:
- fixing #347 : minor documentation fix for the xyz basis

### DIFF
--- a/orbitize/basis.py
+++ b/orbitize/basis.py
@@ -6,6 +6,7 @@ import abc
 from orbitize import priors, kepler
 from scipy.optimize import fsolve
 
+
 class Basis(abc.ABC):
     """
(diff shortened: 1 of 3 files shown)


sblunt:
- add #358 to the list: making documentation for `stellar_or_system_mass` clearer

sblunt:
- add #351: orbitize! was at some point throwing an error when setting `fit_secondary_mass=False` and including companion RVs. I actually couldn't reproduce this bug anymore, but I added a unit test to check that this stays fixed in the future.

semaphoreP (approved):
Looks good

PR #364, opened 2024-03-01
Add to Modifying Priors tutorial

wcroberson:
Added an extra block to the Modifying Priors tutorial notebook, which explains that the system should always be instantiated and modified directly when modifying priors, rather than modifying the sampler or driver object.


sblunt (approved):
Looks good! Thanks for making this change.

sblunt:
The tests were failing because of a v3 issue that has since been corrected. Since this commit does not change any functionality (only documentation), I'm going to merge it without rerunning the tests.

PR #365, opened 2024-03-01
Obs priors

sblunt:
Implementation of the observation-based priors (finally!) 

I added a new basis to match the one derived in Kelly's original paper, and modified the code on my old branch to implement the math correctly. @clarissardoo and I also wrote up a tutorial, and I added some unit tests. I'm running a longer end-to-end fit (using the new tutorial, just running for a goodly number of steps), and I'll post the results from that here so we can visually compare to a published independent implementation.


semaphoreP (approved):
Fix image in docs. and we should be good

sblunt:
Image is displaying for me in the tutorial, and the CI run through the notebooks passed. Not sure why you couldn't see it. I'm going to pull this in but we can fix it later if it's still not working.

PR #373, opened 2024-08-12
Astrometry plotting

DTCupcakes:
Fixed an issue with the astrometry plotting method where coordinate systems for astrometry (ra/dec or sep/pa) were being mislabelled, causing astrometry to be plotted incorrectly.


sblunt:
lol @ me saying "I'll look at this next week" two years ago. It's been an eventful time!

Anyway, thanks so much for making this PR @DTCupcakes ! I think this will be a very useful utility for folks wanting to check that they put their astrometry in correctly. I made a few minor changes to the code itself (shortening it by using existing system attributes and also making it a standalone method, i.e. something that is not always run). I also updated the betaPic_hipIAD end-to-end test to use this as an example. I'll pull it into the v4 branch once the tests pass!

Issue #370, opened 2024-07-15
Result File From Resumed MCMC Fails to Save


HenryDennen:
**Describe the bug**
Trying to resume a previous MCMC fit that used Gaia-Hipparcos, Relative Astrometry, and RV. After initializing a new a MCMC sampler (with the path to the previous hdf5 file passed through the keyword "prev_result_filename") the sampler fails to save the the result (with results.save_results) and gives the below error.
 `FileNotFoundError: [Errno 2] No such file or directory: 'thisisprettyhackysorrylmao'`

**To Reproduce**
1: Click on [this GitHub gist](https://gist.github.com/HenryDennen/fc4ed67b12eef2dd148a735a658cf0f5) 
2. Scroll down to the end of the gist
3. See error

**Expected behavior**
When creating a new MCMC sampler, according to the orbitize! docs, if a filename to an HDF5 containing a .Results object is passed to prev_result_filename, the MCM will restart from where it left off. After running the MCMC sampler for the specified number of burn-in steps and posterior samples, the results should be saved to an HDF5 named "Hip21547_RERUN_results.hdf5". and then print a message saying the results were saved there to that filename.

**Screenshots**
<img width="662" alt="Screen Shot 2024-07-15 at 12 00 10 PM" src="https://github.com/user-attachments/assets/95150e20-039e-49f4-9321-4946e2c3ae73">

**Desktop (please complete the following information):**
 - OS: Rocky **Linux** release 8.9 (Green Obsidian)
 - orbitize! Version3.0.0

**Additional context**
@wbalmer has been helping me out.

wbalmer:
Wanted to bump and say that I encountered this error a while ago but was too lazy to create a new issue at the time (oops!)

sblunt:
Addressed in PR #374

PR #374, opened 2024-08-12
fix for Result File From Resumed MCMC Fails to Save #370

sblunt:
Hey @HenryDennen and @wbalmer, thanks for raising an issue about this! I could reproduce your issue, and the patch in this PR fixes it for me. Could you confirm that this works for you too, then I can pull it into main? Thanks!!

### DIFF
--- a/orbitize/results.py
+++ b/orbitize/results.py
@@ -197,7 +197,7 @@ class Results(object):
         iad_data = hf.get("IAD_datafile")
         if iad_data is not None:
 
-            tmpfile = 'thisisprettyhackysorrylmao'
+            tmpfile = 'tmpfile_OkToDeleteAfterFitFinishes'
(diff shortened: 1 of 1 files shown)


sblunt:
I also extended an existing unit test to cover this situation, since the bug was occurring only after:

- orbit run with Hipparcos IAD 
- orbit fit saved 
- orbit fit loaded into new results file
- that new orbit fit is saved

semaphoreP:
The fix is just to not delete the tmp file? Is that correct?

sblunt:
yup!

wbalmer:
I just confirmed that the fix works for Henry's gist example and some of my own code. Thanks Sarah!

semaphoreP (approved):

PR #375, opened 2024-08-19
fix for 0d orbit masking for xyz basis

sblunt:
This line in system.py was causing a failure when we pass in a single orbit. Caught by the changing basis tutorial! I applied a very simple fix of checking if ecc is a float, then if so making it an array for the logical comparison

### DIFF
--- a/orbitize/system.py
+++ b/orbitize/system.py
@@ -555,7 +555,9 @@ class System(object):
 
         if self.fitting_basis == "XYZ":
             # Find and filter out unbound orbits
-            bad_orbits = np.where(np.logical_or(ecc >= 1.0, ecc < 0.0))[0]
+            if ecc.size == 1:
+                ecc_as_array = np.array([ecc])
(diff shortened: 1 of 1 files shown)


orbitize/system.py
```diff
@@ -555,7 +555,9 @@ def compute_all_orbits(self, params_arr, epochs=None, comp_rebound=False):
 
         if self.fitting_basis == "XYZ":
             # Find and filter out unbound orbits
-            bad_orbits = np.where(np.logical_or(ecc >= 1.0, ecc < 0.0))[0]
+            if ecc.size == 1:
+                ecc_as_array = np.array([ecc])
+            bad_orbits = np.where(np.logical_or(ecc_as_array >= 1.0, ecc_as_array < 0.0))[0]
```
semaphoreP, line 560: what happens when `ecc.size != 1`? I don't see `ecc_as_array` to be defined in that case.

sblunt:
Oops, forgot I had opened this PR and made an (incomplete) fix as Jason pointed out above. This issue was resolved by a different PR #388. Closing this.

PR #379, opened 2024-09-24
Nested sampling with MultiNest

tomasstolker:
This PR adds the `MultiNest` class to the `sampler` module. MultiNest supports the use of MPI, so it can be executed as e.g. `mpirun -n 100 python run_orbitize.py` if 100 CPUs are available.

Also made some minor modifications in `NestedSampler` that uses Dynesty. I added the `nlive` parameter, which is used by the static sampler. I removed a `ValueError`, which seemed not needed since `pfrac` is not used when `static=True`.

The unit tests do not actually call `run_sampler`, since MultiNest needs to be manually installed.

Let me know if you have any comments/feedback!


sblunt:
Merging this branch into develop, and will continue the conversation over there!

Issue #372, opened 2024-07-31
Possible bug in Hipparcos calculations regarding decimal years and leap years


sefffal:
**Describe the bug**
I'm working on hipparcos modelling in my own code and ran into this problem. I thought I'd check how orbitize! does it, but I think we might both have the same error.

The hipparcos IAD residual data gives times in units of "1991.25 + year". In the hipparcos file, currently the code does:

```
times = iad[1] + 1991.25
# later...
epochs = Time(times, format="decimalyear")
```

The problem is I don't think that the offsets computed as 1991.25 + fractional_year should be considered a decimal year time. This is an issue only for scans between 1992 and 1993, because 1992 is a leap year with an extra day (366 days instead of 365).

cc @semaphoreP

sefffal:
I think it would be more correct to convert 1991.25 from decimal year to MJD, and then add offsets multiplied by one Julian year (365.25). But I'm not positive.

sefffal:
One more note: I believe the Hipparcos catalog epoch is actually "J1991.25", which isn't technically a decimal year either:

```
Time("J1991.25", format="jyear_str").decimalyear
```
Gives an answer that is 7 hours different than Time(1991.25, format="decimal year").

sblunt:
Hey! Thanks so much for pointing this out (and sorry it's taken me so long to dig into it). I'm digging around for more info, but so far I'll share:

- [This website](https://heasarc.gsfc.nasa.gov/w3browse/all/hipparcos.html) gives a conversion from J1991.25 to JD which matches your third comment suggestion. So that's quite good evidence that at least the J1991.25 epoch should be read in the way you suggest!
- It would make sense to me that the delta_epochs are actually in Julian year rather than decimal year. I'd like to find some piece of documentation that says that to be sure. I'm going to keep googling a bit. 

Let me know if you've found any more info!

sblunt:
Whoops, just realized you said exactly that in your first comment. Sorry for restating the obvious!

sblunt:
More random info dumping of sources: on page 288 of the [2007 Hipparcos 2 reduction](https://link.springer.com/book/10.1007/978-1-4020-6342-8) it mentions "where t is measured in years
of 365.25 days, counted from 1991.25." The context is a little different, but that's pretty good evidence for me that this is all defined in Julian years as you suggested! I'm going to make that change in my code and see how much it affects some of our tests.

sblunt:
[...]

PR #382, opened 2024-10-08
Replace decimal year conversions with julian year conversions.

sblunt:
See issue #372. This doesn't have any big impact on the Hipparcos IAD refitting, but maybe it affects things on longer timescales (i.e. hipparcos-gaia). Will continue to investigate.


orbitize/hipparcos.py
```diff
@@ -17,29 +16,28 @@ class PMPlx_Motion(object):
...
-            taken with GRAVITY, as well as a Hipparcos-derived position (at epoch 1991.25),
-            alphadec0_epoch should be 1991.25, and you should pass in absolute astrometry
+            taken with GRAVITY, as well as a Hipparcos-derived position (at epoch J1991.25),
+            alphadec0_epoch should be J1991.25, and you should pass in absolute astrometry
             in terms of mas *offset* from the Hipparcos catalog position, with propagated
             errors of your measurement and the Hipparcos measurement.
     """
 
-    def __init__(self, epochs_mjd, alpha0, delta0, alphadec0_epoch=1991.25):
+    def __init__(self, epochs_mjd, alpha0, delta0, alphadec0_epoch=Time("J1991.25", format="jyear_str").jyear):
```
semaphoreP: probably shouldn't initialize a keyword with a non-primitive type for python reasons.

semaphoreP (commented):
I think it generally looks good to me. Did you have a chance to run an HGCA orbit fit to see if it matches previous results?

orbitize/hipparcos.py
sblunt: good call, thanks!

PR #384, opened 2025-02-04
Add brigtness calc to comp all orbits

Farrmol:


orbitize/kepler.py
```diff
@@ -44,6 +44,42 @@ def tau_to_manom(date, sma, mtot, tau, tau_ref_epoch):
 
     return mean_anom
 
+def times2trueanom_and_eccanom(
```
semaphoreP, line 47: missing docstring

orbitize/system.py
```diff
@@ -489,16 +497,33 @@ def compute_all_orbits(self, params_arr, epochs=None, comp_rebound=False):
...
+
+                R = (sma*(1-ecc**2))/(1+ecc*np.cos(tanom))
+        
+                z = (R)*(-np.cos(argp)*np.sin(inc)*np.sin(tanom)-np.cos(tanom)*np.sin(inc)*np.sin(argp))
+                
+                B = np.arctan2(-R, z)+ np.pi
+        
+                alpha = (1/np.pi)*(np.sin(B)+(np.pi-B)*np.cos(B))
+
+                albedo = 0.5 # NOTE: we're only fitting relative changes in brightness, so the actual value of albedo doesn't matter
```
semaphoreP, line 516: if we're fitting relative changes, how will that work? Are we going to analytically marginalize over some linear scale factor term in the future in lnlike?

orbitize/system.py
```diff
@@ -489,16 +497,33 @@ def compute_all_orbits(self, params_arr, epochs=None, comp_rebound=False):
...
+                brightness = albedo*alpha/R**2
```
semaphoreP, line 517: maybe not for this PR, but we should make this part of code customizable: we can replace it with a user-provided function or class, so that we can handle custom and non-Lambertian models. We could refactor this into a separate function here or make an issue to do it.

orbitize/system.py
semaphoreP, line 373: update docstring to add fourth return element. I think we also (maybe separately) need to update the orbitize release notes about this breaking function signature change for `compute_all_orbits()`. It may break people's code that use it.

tests/test_brightness.py
```diff
@@ -0,0 +1,92 @@
...
+    brightness_data = data_table[data_table['quant_type'] == 'brightness']
+    assert np.all(brightness_data['quant1'].value == [0.9, 0.6])
+    assert np.all(brightness_data['quant1_err'].value == [0.1, 0.2])
+    assert np.all(np.isnan(brightness_data['quant2'].value))
+    assert np.all(np.isnan(brightness_data['quant2_err'].value))
+
+
+def test_compute_posteriors():
+    """
+    Test that a short mcmc runs to completion
```
semaphoreP: is this test just to show that adding brightness to the csv doesn't break anything? are we actually fitting for it right now?

semaphoreP (commented):
Nice, I like it. I think we have some documentation updates (especially since this is a breaking change to `compute_all_orbits()`. I also had some questions below about how this will work in the future.

tests/test_brightness.py
semaphoreP: Update the docstring to explain the purpose of the test

orbitize/system.py
sblunt: Great question. Noting for posterity here our offline conversation that it indeed makes sense to fit a linear scale factor. I just raised issue #416 in response to this comment.

orbitize/system.py
sblunt: Agreed! I raised issue https://github.com/sblunt/orbitize/issues/417 to keep track. Let's do it in a future PR.

orbitize/system.py
sblunt: Good call. I'll update the release notes after I pull into main, and will be sure to note this is a breaking change.

sblunt:
Thanks for the review @semaphoreP! I'm going to pull the changes to main and release a new version now. The CI tests on this branch are broken because they're using an old version of the coveralls action. They should be fixed when pulled into main.

PR #389, opened 2025-07-17
modernize orbitize install & fix ci tests

sblunt:
Not everything is working yet but I want to test the automatic notebook running workflow which triggers on pr to main


sblunt:
All right, assuming tests pass, I think everything is working now. Main changes I made:
- moved most of the pip setup code from setup.py to pyproject.toml. It seems like pyproject.toml still doesn't have a clean functionality to compile cython code, so I'll keep that part and a setup.py for now
- debugged two tests (one jupyter notebook and one multiprocessing one that was failing on my mac)

PR #406, opened 2026-05-27
Nautilus

### LINKED ISSUE #387: nautilus
Sven Heydenreich recommends checking out this sampler, which is getting popular in cosmology: https://github.com/johannesulf/nautilus

eshelDror:
Add basic functionality for the nautilus nested sampler https://github.com/johannesulf/nautilus
Add to requirements

### DIFF
--- a/orbitize/sampler.py
+++ b/orbitize/sampler.py
@@ -7,6 +7,8 @@ from astropy.time import Time
 
 import dynesty
 
+import nautilus
+
 import emcee
 import ptemcee
 import multiprocessing as mp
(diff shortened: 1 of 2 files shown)


sblunt (changes requested):
Hey Eshel et al! Nice work on this PR. I have two minor modification requests:

1. I think it would be nice for the NautilusSampler object to inherit from the dynesty-based NestedSampler object so that we don't have the repeated prior transform code. Better yet, we could have a BaseNestedSampler class that both NestedSampler and NautilusSampler classes inherit from.

2. Rather than using strong-typing, please follow the docstring format of other orbitize! classes/functions. This makes sure that our docs are built correctly (see e.g. https://orbitize.readthedocs.io/en/latest/sampler.html)

sblunt:
Oh actually, one other thought-- in the current main branch, there is a new implementation of multinest, also a nested sampling algorithm. It would be good to merge the current version of main in here and implement the base class for all three nested sampling algorithms.

sblunt:
Moved to #408

Issue #378, opened 2024-09-23
Setting elegant default plotting settings


gabrielastro:
Congratulations on your orbitize! v3 paper on arXiv! Maybe v3 would be a good point to set pretty default settings for the plots :)? My main recommendations are:

1. Use mirror ticks (i.e., ticks on the right and top): it also makes reading off values so much easier, especially for curves close to those edges.

2. Use minor ticks: also very useful for reading off values, and elegant.

3. Use a bigger font size: in a paper, it is not practical to have to zoom in just to read the axes and/or the values… The trick is here actually to make the figures intrinsically _smaller_, so that the elements are bigger relative to the plot and thus the page.

4. This one is more subjective but a prettier font than the default one would be nice. There are many options that are elegant and legible. I am not a `python` programmer but from what I can tell, looking at the source of [species](https://github.com/tomasstolker/species), it is as simple as doing:
```
import matplotlib.pyplot as plt
…
plt.rcParams["font.family"] = "serif"
plt.rcParams["mathtext.fontset"] = "dejavuserif"
```
I do not mean that you need to take same font—even though you could, of course. Just make sure that the text and math fonts look similar; otherwise, one gets ugly mixes e.g. within a label. If you want to go for the classical Hershey Complex font, there is [smplotlib](https://github.com/AstroJacobLi/smplotlib) (as simple as: `import smplotlib; matplotlib.rcParams['axes.linewidth']=1.` because the default is (currently) too thin), and in general there is more of this kind of advice under https://turnermoni.ca/python3.html. Make sure also to use a plot border thickness that goes well with the font.

**Describe alternatives you've considered**
Non-prettiness is not an alternative :wink:.

Thanks for considering this! It would make your code even more attractive to readers who are not users… yet. I do not know how you can set the settings centrally to give the plots a consistent look but maybe there is an easy solution!

sblunt:
See comment in #388

PR #415, opened 2026-07-15
Plot refactor

eshelDror:


orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
 
 
-# TODO: deprecatation warning for plots in results
-
-# define modified color map for default use in orbit plots
 cmap = mpl.cm.Purples_r
 cmap = colors.LinearSegmentedColormap.from_list(
     "trunc({n},{a:.2f},{b:.2f})".format(n=cmap.name, a=0.0, b=0.7),
     cmap(np.linspace(0.0, 0.7, 1000)),
 )
```
sblunt, line 27: Do we need to define cmap here and below? Can we reduce it to just one def, either at the top here or in the Plotter class?

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+    )
+    POSSIBLE_CBAR_PARAMS = ["sma", "ecc", "inc", "aop" "pan", "tau", "plx", "m0", "m1"]
+    # colour/shape scheme scheme for rv data points
+    RV_COLORS = ("#0496FF", "#372554", "#FF1053", "#3A7CA5", "#143109")
+    RV_SYMBOLS = ("o", "^", "v", "s")
+
+    def __init__(
+        self,
+        results,
+        object_to_plot=1, # TODO: Support multiplanet
```
sblunt: Noting that we should resolve this todo before merge (I'm sure this is already on your radar).

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+        num_orbits_to_plot=100,
+        num_epochs_to_plot=100,
+        cbar_param="Epoch [year]",
+        rv_time_series=False,
+        rv_time_series2=False
+    ):
+        self.results = results
+        self.system = results.system
+
+        if default_cmap is not None:
```
sblunt: Make sure the logic works here to define the colormap as we expect

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+        time_format (str): time format for `start` and `end` such as "mjd" or others
+            from astropy.time.Time.FORMATS (default: 'decimalyear')
+        default_cmap (matplotlib.cm.ColorMap): color map to use for making orbit tracks
+            (default: modified Purples_r)
+        num_orbits_to_plot (int): number of orbits to plot (default: 100)
+        num_epochs_to_plot (int): number of points to plot per orbit (default: 100)
+        cbar_param (string): options are the following: 'Epoch [year]', 'sma1', 'ecc1', 'inc1', 'aop1',
+            'pan1', 'tau1', 'plx', 'm0', 'm1'. Number can be switched out. (default: 'Epoch [year]').
+        rv_time_series (Boolean): if fitting for secondary mass using MCMC for rv fitting,
+            calculate rv_time_series of the primary (object 0) (default: False)
```
sblunt, line 47: let's say "plot rv time series of the primary" instead of calculate, and same for rv_time_series2. In rv_time_series2, we should also specify this is the relative RV (i.e. RV of secondary - RV of primary)

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+        rv_time_series2=False
+    ):
+        self.results = results
+        self.system = results.system
+
+        if default_cmap is not None:
+            self.cmap = default_cmap
+        
+        if start is None:
+           start = getattr(Time(np.min(self.system.data_table['epoch'])-365*3, format="mjd"), time_format)
```
sblunt, line 94: I'm tempted to automatically set the start and end times for each dataset being plotted, and asking the user to manually override these after the Figure object is returned. If we make this change, we should include an example of manually setting axis limits in the plotting tutorial. Also, if there is no data, but the user still wants to plot the prediction, we can use the "first epoch (of any kind) - 3 years" as the start date, and analagous for end date.

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+        object_to_plot=None,
+        start=None,
+        end=None,
+        time_format=None, 
+        num_orbits_to_plot=None,
+        num_epochs_to_plot=None,
+        cbar_param=None,
+        rv_time_series=None,
+        rv_time_series2=None,
+    ):
```
sblunt, line 112: add docstring

orbitize/plot.py
```diff
@@ -186,6 +1527,119 @@ def plot_orbits(
...
+    )
+    return plotter.plot_propermotion(
+        periods_to_plot=periods_to_plot,
+        alpha=alpha,
+        show_colorbar=show_colorbar,
+        tight_layout=tight_layout
+    )
+
+
+def legacy_plot_orbits(
```
sblunt: Decided you can remove the legacy code. It will be in the git history if we want to refer back to it.

orbitize/plot.py
```diff
@@ -186,6 +1527,119 @@ def plot_orbits(
     primary_instrument_name=None,
     fontsize=20,
     fig=None,
+):
+    # TODO: Improve message
+    print("The orbitize.plot.plot_orbits function is deprecated. Instead, initialize an orbitize.plot.Plotter and use the plot_orbits method.")
```
sblunt: Very nice. Just writing down what we talked about over zoom for posterity that we should raise a DeprecationWarning (or similar) here

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+                self.start, self.num_orbits_to_plot, self.num_epochs_to_plot, self.object_to_plot, self.standard_post, self.end)
+            self.cbar_param_arr, self.norm, self.norm_yr = self._create_cbar(self.cbar_param, self.epochs, self.standard_post)
+            if self.rv_time_series:
+                self.rv_data, self.insts, self.gams, self.labels, self.gam_idx, self.rv_inst_inds, self.sig_idx = self._calc_rv(object=0)
+            if self.rv_time_series2:
+                self.rv_data2, self.insts2, self.gams2, self.labels2, self.gam_idx2, self.rv_inst_inds2, self.sig_idx2 = self._calc_rv(object=1)
+            self.sep_data, self.sep_err, self.pa_data, self.pa_err, self.ra_data, self.ra_err, self.dec_data, self.dec_err = self._calc_seppa_radec(self.data)
+            self.astr_raoff, self.astr_deoff, self.astr_vz, self.astr_epochs = self._calc_astr_orbits(self.standard_post, self.num_orbits_to_plot, self.object_to_plot)
+
+    def _calc_seppa_radec(self, all_data):
```
sblunt: add docstrings for all these private functions

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+        data = all_data[all_data["quant_type"]!="rv"]
```
sblunt: can you calculate all of these data type masks outside of these private methods? It would be nice to see them in one place (maybe in set_params above)

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+        astr_epochs = self.data["epoch"][astr_inds]
+        num_astr_epochs = len(astr_epochs)
+
+        raoff = np.zeros((num_orbits_to_plot, num_astr_epochs))
+        deoff = np.zeros((num_orbits_to_plot, num_astr_epochs))
+        vz = np.zeros((num_orbits_to_plot, num_astr_epochs))
+        # TODO: vectorize
+        for i in np.arange(num_orbits_to_plot):
+            # Calculate ra/dec offsets for all epochs of this orbit
+            raoff0, deoff0, vz0 = self.system.compute_all_orbits(
```
sblunt: note we just changed the return type of compute_all_orbits in v3.4.0, so you will need to add raoff0, deoff0, vz0, _ here (and elsewhere this method is called)

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+                astr_epochs
+            )
+
+            raoff[i, :] = raoff0[:, object_to_plot, 0]
+            deoff[i, :] = deoff0[:, object_to_plot, 0]
+            vz[i, :] = vz0[:, object_to_plot, 0]
+        
+        return raoff, deoff, vz, astr_epochs
+
+    def _calc_rv(self, object):
```
sblunt: we should change object to something else since object is a special variable in python. It should work fine as is but might be confusing.

orbitize/plot.py
sblunt: Actually, I disagree with myself now that I went through the whole refactor. I like it as is.

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+        # indices corresponding to each instrument in the datafile
+        inds = {}
+        for i in range(len(insts)):
+            inds[insts[i]] = np.where( # include encode for backwards compatibility
+                (rv_data["instrument"] == insts[i].encode()) | (rv_data["instrument"] == insts[i])
+            )[0]
+        return rv_data, insts, gams, labels, gam_idx, inds, sig_idx
+
+    def _get_standard_post(self, num_orbits_to_plot):
+        # TODO: Replace random with results.downsample
```
sblunt, line 517: good call to replace with results.downsample!

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+        square_plot=True,
+        show_colorbar=True,
+        cmap=None,
+        sep_pa_color="lightgrey",
+        mod180=False,
+        plot_astrometry=True,
+        plot_astrometry_insts=False,
+        rv_time_series=False,
+        rv_time_series2=False,
+        # plot_errorbars=True,
```
sblunt: remove commented out code

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+            )
+
+        if (rv_time_series and not self.rv_time_series) or (rv_time_series2 and not self.rv_time_series2):
+            self.set_params(rv_time_series=rv_time_series, rv_time_series2=rv_time_series2)
+
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", ErfaWarning)
+
+            # Before starting to plot rv data, make sure rv data exists:
+            rv_indices = np.where(self.data["quant_type"] == "rv")
```
sblunt: Can we replace this bit of code with the self.rv_data variable you defined above?

sblunt: Actually, I think we should raise this warning in self.set_params.

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+                if (rv_time_series) and (rv_time_series2):
+                    ax = plt.subplot2grid((4, 16), (0, 0), rowspan=2, colspan=6)
+                elif (rv_time_series) and (not rv_time_series2):
+                    ax = plt.subplot2grid((3, 16), (0, 0), rowspan=2, colspan=6)
+                elif (not rv_time_series) and (rv_time_series):
+                    ax = plt.subplot2grid((3, 16), (0, 0), rowspan=2, colspan=6)
+                else:
+                    ax = plt.subplot2grid((2, 14), (0, 0), rowspan=2, colspan=6)
+
+            astr_inds = np.where((~np.isnan(self.data["quant1"])) & (~np.isnan(self.data["quant2"])))
```
sblunt: Let's use the variables you defined above rather than recalculating this mapping here.

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+
+            # For plotting different astrometry instruments
+            if plot_astrometry_insts:
+                astr_colors = ("#FF7F11", "#11FFE3", "#14FF11", "#7A11FF", "#FF1919")
```
sblunt: define these in the same place you define the RV data colors and symbols above.

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+
+            for ax2 in fig.get_axes():
+                ax2.tick_params(axis="both", labelsize=15)
+                ax2.minorticks_on()
+
+        fig.tight_layout()
+        
+        return fig
+
+    def _plot_full_orbits(self, ax, plot_astrometry, square_plot, fontsize, cmap, plot_astrometry_insts, astr_insts=None, astr_inst_inds=None, astr_colors=None, astr_symbols=None):
```
sblunt: I'd suggest putting all the private methods together in the beginning

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+        m1 = self.standard_post[
+            :, self.results.standard_param_idx["m{}".format(self.object_to_plot)]
+        ]
+        mtot = m0 + m1
+        for i in np.arange(self.num_orbits_to_plot):
+            if rv_time_series:
+                plt.sca(ax3)
+
+                # scale back to primary RV semi amplitude
+                vz0 = self.vz1[i] * (-(mtot[i] - m0[i]) / np.median(m0[i])) # TODO: vectorize
```
sblunt: we should use the output of calc_all_orbits here rather than doing the math ourselves

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+                    color=sep_pa_color,
+                )
+
+            if rv_time_series2:
+                if rv_time_series:
+                    plt.sca(ax4)
+                else:
+                    plt.sca(ax3)
+
+                # scale back to primary RV semi amplitude
```
sblunt: Looks like this comment isn't relevant anymore; we can remove it

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+            )
+
+            plt.sca(ax1)
+            plt.plot(yr_epochs, seps, color=sep_pa_color)
+
+            plt.sca(ax2)
+            plt.plot(yr_epochs, pas, color=sep_pa_color)
+    
+    def _plot_rv_data(self, ax3, ax4, rv_time_series, rv_time_series2, sep_pa_color):
+        # plot RV orbits here
```
sblunt: let's call these _plot_rv_model, _plot_sep_pa_model, etc, since the data is being plotted in a separate place

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+
+
+            # get rvs and plot them
+            for i, name in enumerate(self.rv_inst_inds.keys()):
+                inst_data = self.rv_data[self.rv_inst_inds[name]]
+                rvs = inst_data["quant1"]
+                epochs = inst_data["epoch"]
+                epochs = Time(epochs, format="mjd").decimalyear
+                inst_err = inst_data["quant1_err"]
+                gam_err2 = np.sqrt(np.square(inst_err)+np.square(med_sigma[i]))
```
sblunt: these (gam_err1, gam_err2, gam_err_2) could use more informative variable names

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+                gam_err_1 = np.sqrt(gam_err2)
+                gam_err_2 = np.sqrt(gam_err2 + np.square(stddev_ga[i]))
+                c = next(ax3_colors)
+                plt.scatter(
+                    epochs,
+                    rvs-med_ga[i],
+                    s=30,
+                    marker=next(ax3_symbols),
+                    c=c,
+                    label=name,
```
sblunt: if the name is "defrv" (our default when the user doesn't provide an instrument name) we should make the label something more informative like "RV data"

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+
+            # get rvs and plot them
+            for i, name in enumerate(self.rv_inst_inds.keys()):
+                inst_data = self.rv_data[self.rv_inst_inds[name]]
+                rvs = inst_data["quant1"]
+                epochs = inst_data["epoch"]
+                epochs = Time(epochs, format="mjd").decimalyear
+                inst_err = inst_data["quant1_err"]
+                gam_err2 = np.sqrt(np.square(inst_err)+np.square(med_sigma[i]))
+                gam_err_1 = np.sqrt(gam_err2)
```
sblunt: subtle point here-- each orbit may have its own fitted value of jitter (sigma). We want to use that value, rather than the median across all orbits.

sblunt: Also, I think gam_err_1 is unneeded. Below, gam_err_2 should be np.sqrt(gam_err2**2 + stddev_ga[i]**2)

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+                    marker=next(ax3_symbols),
+                    c=c,
+                    label=name2,
+                    zorder=5,
+                )
+                if separate_gamma_err:
+                    plt.errorbar(
+                        x=epochs2,
+                        y=rvs2-med_ga2[i],
+                        yerr=stddev_ga2[i],
```
sblunt: Nice job working through all the differences here!

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+                        ecolor=c,
+                        zorder=5,
+                        ls="none",
+                        label="{} gamma err".format(name2)
```
sblunt: I'd probably suggest writing these captions out in latex to make them as clear as possible: std(rv_{inst name} - gamma_{inst name}); sigma_{inst name}**2 + sigma**2 (and same for the other case below)

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+            ``matplotlib.pyplot.Figure``: the residual plots
+
+        Refactored by Eshel Dror, 2026        
+    
+        """
+        raoff = self.astr_raoff
+        deoff = self.astr_deoff
+        seps = []
+        pas = []
+        raoff_100 = self.raoff1
```
sblunt: use your defined variables throughout rather than redefining them (for clarity)

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+
+            raoff2, deoff2 = raoff_100[i], deoff_100[i]
+
+            seps1, pas1 = orbitize.system.radec2seppa(raoff0, deoff0, mod180=mod180)
+            # seps1 = []
+            # pas1 = []
+
+            # for j in range(len(astr_epochs)):
+
+            #     seps0, pas0 = orbitize.system.radec2seppa(raoff[i][j], deoff[i][j], mod180=mod180)
```
sblunt: remove commented out code

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+        Plots sep/PA residuals for a set of orbits
+
+        Args:
+            sep_pa_color (string): any valid matplotlib color string, used to set the
+                color of the orbit tracks in the Sep/PA panels (default: 'lightgrey').
+            mod180 (Bool): if True, PA will be plotted in range [180, 540]. Useful for plotting short
+                arcs with PAs that cross 360 deg during observations (default: False)
+
+        Return:
+            ``matplotlib.pyplot.Figure``: the residual plots
```
sblunt, line 1520: Let's use the same logic for plotting residuals as we used above for plotting the gammas for RVs. Offer two options for plotting the uncertainties, exactly the same way you did above.

orbitize/plot.py
```diff
@@ -19,17 +19,1328 @@
...
+            for i in np.arange(self.num_orbits_to_plot):
+                epoch_in_yr = Time(epochs[i, :], format="mjd").decimalyear
+                # masses (in same units, solar)
+                m_b = self.standard_post[:, self.results.param_idx["m1"]][i]
+                m_a = self.standard_post[:, self.results.param_idx["m0"]][i]
+                # dt
+                timestep = epoch_in_yr[1] - epoch_in_yr[0]
+                # dra/dt and ddec/dt
+                ddec_b = np.gradient(deoff[i, :], timestep)  # in mas/yr
+                dec_b_radian = (
```
sblunt, line 1750: let's use astropy.units to do these conversions for clarity

sblunt:
Looks great! Excellent job @eshelDror. A few small overall comments:
1. Please push the changes to the notebooks without outputs (i.e. run. "clear output" before pushing). That makes the diff a little cleaner. The plots will be autogenerated by nbconvert when the documentation is building, and you'll see the plots on orbitize.info.
2. In general, it would be very good to use a consistent set of mask variables (e.g. self.rv_data) throughout, rather than redefining them in each subfunction. This is definitely orbitize's fault, not yours, but since you're refactoring the plot code, I think you should be the one to fix it. :D I left a few in-line comments with suggested places to make these changes, but I don't think I got all of them.

orbitize/plot.py
```diff
@@ -162,8 +1476,35 @@ def plot_corner(results, param_list=None, **corner_kwargs):
         corner_kwargs["labels"] = reduced_labels_list
 
     figure = corner.corner(samples, **corner_kwargs)
-    return figure
 
+    if plot_priors:
+        axes = figure.axes
+        num_params = len(param_indices)
```
sblunt, line 2044: I think you can avoid much of this normalization code by passing hist2d_kwargs = {"density":True} into corner.corner.

orbitize/plot.py
```diff
@@ -162,8 +1476,35 @@ def plot_corner(results, param_list=None, **corner_kwargs):
...
+        sample_size = samples.shape[0]
+        for i, param_i in enumerate(param_indices):
+            prior = results.system.sys_priors[param_i]
+            if not hasattr(prior, "compute_lnprob"):
+                continue
```
sblunt, line 2048: This should never happen; Prior objects by definition (by means of the abstract base class) must have a compute_lnprob method, so I think you can safely remove this check

sblunt:
Also, see aesthetic suggestions in #378 and please implement the ones you like! I think the main things I'd like to see here based on those suggestions are:
1. by default using circles instead of stars for astrometric data
2. using minor ticks and mirror ticks on all panels
3. using a (slightly) bigger font size
4. I also think the default red astrometry is a little jarring. Play around with the default color scheme and see what you can figure out that you like!

sblunt:
Noting for myself that the Gaussian prior plotting fix in this PR should also be a separate changelog item

sblunt:
Ok, I think the code looks great. I have a few minor remaining aesthetic requests:

1. HGCA plot-- it's hard to see the errorbars with the current color scheme. Can we mke the error bars black? I like the black outline around the markers. Can we do that for data points for the astrometric plots too? Finally, I like the blue and purple-- can we change the orange points in the astrometric plot to the light blue with black outlines?
2. ALso for the HGCA plot, can you add plt.tight_layout() to make the dimensions look a little cleaner?
3. I'm wondering if there is an easy way to communicate that the orange lines are the priors. Can you add a legend in the top panel that labels "posterior," "prior," eg?
4. I like the description of the Plotter class you added to the plotting tutorial!
5. In the residuals plot, can you label "model standard deviation" similarly to the way you label the RV errors induced by the gamma posterior width? ie in latex; std(data-model). I also think the colors here are a little hard to see. Can we do 1. rebeccapurple instead of purple, and the same blue instead of orange (and with the data points outlined in black). I also think making one error thicker and with a lower alpha will be helpful for the visual here.
6. RV plots-- let's keep the colors of the data points consistent with the colors and styles in the astrometry points here, and the error bars the same color scheme as the astrometric residuals plots. I thinkt he thicker error bar should be 2 or 3 times thicker and also have a lower alpha. Minor point: $std(\gamma)$ should be std($\gamma$) so the std isn't italicized. You can also do $\mathrm{std}(\gamma)$.

sblunt:
I also think the multiplanet plots look great!

PR #419, opened 2026-08-04
Remove gpu code

sblunt:
Technically this is a breaking change and I should be more careful about throwing deprecation warnings, but I don't think anyone is actually using this functionality, and it was basically invisible to users.


sblunt:
Accidentally created branch for this PR not from main; closing & starting over

PR #427, opened 2026-08-25
Fix nbody units

eshelDror:
Convert from AU/year to km/s
Before:
K3_regular_orbittracks
K3_regular_differences

After:
K3_nbody_fix_orbittracks
K3_nbody_fix_differences


sblunt:
Excellent catch. Thanks!

sblunt (approved):

PR #428, opened 2026-08-26
Remove unnecessary astroquery imports

eshelDror:
Only import modules which require `astroquery` when necessary to not get stuck when the internet/archive are very slow. Has been tested with the Gaia archive being online, slow, and down.

### DIFF
--- a/orbitize/results.py
+++ b/orbitize/results.py
@@ -5,9 +5,7 @@ import os
 import astropy.table as table
 
 import orbitize.system
-import orbitize.basis
 import orbitize.plot
-import orbitize.gaia, orbitize.hipparcos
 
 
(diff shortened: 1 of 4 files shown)


sblunt:
Great work on this. The complicatedness of this has convinced me it is not worth having the online functionality in the codebase itself. Let's just get rid of the astroquery functionality altogether in the main code base. We can keep it in the tutorials so that users can still benefit. We should do this for the Gaia astroquery and the HGCA download. I think for the HGCA download, we can run the tests with a portion of the HGCA file (a large file), and make that available in the example_data directory. I think those are all the online parts of the code... would be good to run the tests purely offline to check.

PR #430, opened 2026-09-01
Optimize Kepler Calculations in C

eshelDror:
The full solver is implemented in C
`kepler.calc_orbit` now also returns a `tanom` array which is used in brightness calculations so it does not have to be recalculated
Breaking changes to `kepler.calc_orbit` which returns an extra result and `_calc_ecc_anom` which is now `calc_ecc_anom`
20,000 MCMC steps on 1 core now take 44 seconds rather than 545


tests/test_kepler_solver.py
```diff
@@ -11,37 +11,38 @@
 from orbitize import cuda_ext
 from orbitize import cext
 
-threshold = 1e-5
+solver_threshold = 2e-5
+ecc_threshold = 1e-5
```
sblunt: Can you add a quick inline comment here motivating why you chose these numbers (recognizing that when we did this originally, we did not do this). Something like "threshold determined empirically by comparison with rebound."

orbitize/_kepler.pyx
sblunt, line 1: Let's rename all these _kepler.pyx rather than adding the 3

orbitize/kepler.py
```diff
@@ -5,25 +5,25 @@
...
 
-if cuda_ext:
-    # Configure GPU context for CUDA accelerated compute
-    from orbitize import gpu_context
-    kep_gpu_ctx = gpu_context.gpu_context()
+if cext:
+    from . import _kepler3
 
-def tau_to_manom(date, sma, mtot, tau, tau_ref_epoch):
+def tau_to_manom_py(date, sma, mtot, tau, tau_ref_epoch):
```
sblunt: Why do you need to rename this? Can we use the same scheme we were using before and import C versions of functions if cext is set?

orbitize/kepler.py
```diff
@@ -5,25 +5,25 @@
 import astropy.units as u
 import astropy.constants as consts
 
-from orbitize import cuda_ext, cext
+from orbitize import cext
 
-if cext:
-    from . import _kepler
+MAX_ITER = 10
```
sblunt: Since these are defaults that the user can change, I'd rather keep them in the function definition rather than as global variables. It would be confusing for the user if orbitize.kepler.TAU_REF_EPOCH = 58849 when they have set tau_ref_epoch = 0, for example.

orbitize/kepler.py
```diff
@@ -5,25 +5,25 @@
...
+def tau_to_manom_py(date, sma, mtot, tau, tau_ref_epoch):
     """
     Gets the mean anomlay
     
     Args:
         date (float or np.array): MJD
         sma (float): semi major axis (AU)
         mtot (float): total mass (M_sun)
-        tau (float): epoch of periastron, in units of the orbital period
+        tau (float): fraction of the orbit which is elapsed at the epoch of
```
sblunt, line 25: Nice, thanks for making the docstring clearer here!

orbitize/kepler.py
sblunt: Actually, I guess we could keep the structure here but just changes the names of these variables. It is nice to not have magic numbers in the functions. Perhaps we can just call them DEFAULT_MAX_ITER and DEFAULT_TAU_REF_EPOCH e.g.

orbitize/kepler.py
```diff
@@ -83,32 +81,96 @@ def times2trueanom_and_eccanom(
...
-    tanom = 2.*np.arctan(np.sqrt((1.0 + ecc)/(1.0 - ecc))*np.tan(0.5*eanom))
+    tanom = (2.*np.arctan(np.sqrt((1.0 + ecc)/(1.0 - ecc))*np.tan(0.5*eanom)))
 
-    return tanom, eanom
+    return np.squeeze(tanom)[()], np.squeeze(eanom)[()]
 
+def make_array(obj):
+    """
+    Make a scalar or 1-dimensional numpy array into a c-compatible 1-dimensional numpy array or floats
+    """
```
sblunt, line 112: Add description of obj to arg list in the docstring

orbitize/kepler.py
sblunt: I see, later on I got to the part where you keep the original function names but rename the c versions and python versions. I like that. I think it would be good to keep that structure for tau_to_manom too so you don't lose backwards compatibility with that function (I think you end up changing the name in one of the tests)

orbitize/kepler.py
```diff
@@ -308,57 +416,25 @@ def _newton_solver(manom, ecc, tolerance=1e-9, max_iter=100, eanom0=None):
     niter = 0
     while ((ind[0].size > 0) and (niter <= max_iter)):
         eanom[ind] -= diff[ind]
-        # If it hasn't converged after half the iterations are done, try starting from pi
```
sblunt, line 311: Why did you end up removing this logic? Does it happen very rarely?

orbitize/_kepler.pyx
sblunt: That will make it easier for me to review too so I can see what changed in the c file

orbitize/system.py
```diff
@@ -502,9 +502,6 @@ def compute_all_orbits(self, params_arr, epochs=None, comp_rebound=False):
                     tau_ref_epoch=self.tau_ref_epoch,
                 )
 
-                tanom, eanom = kepler.times2trueanom_and_eccanom(sma, epochs, mtot, ecc, tau, tau_ref_epoch=self.tau_ref_epoch)
```
sblunt, line 506: Very nice catch here!

orbitize/kepler.py
sblunt: Not in the tests, it's in basis.py

orbitize/priors.py
sblunt, line 1: Very nice optimizations here too.

sblunt:
Looks awesome, and I think the results speak for themselves! I left a few minor comments.
I'll take another look after the renaming so I can compare the C module more easily, but I think it's pretty good to go. Noting to myself here that this PR also includes optimizations of the prior computations, which I think should be a separate item in the changelog.

PR #431, opened 2026-09-01
Make tests consistent

eshelDror:
Reset random seed before each test
Does not work for OFTI tests because `OFTI._sampler_process` resets the seed

### DIFF
--- a/tests/test_multiplanet.py
+++ b/tests/test_multiplanet.py
@@ -224,7 +224,6 @@ def test_fit_selfconsist():
     assert samp.num_params == 6
 
     # start walkers near the true location for the orbital parameters
-    np.random.seed(123)
     # planet b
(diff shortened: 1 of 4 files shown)


sblunt:
Nice. Why don't we add an argument to `OFTI._sampler_process` that accepts an input seed, and sets it if it is not given? That would let us remove all the randomness from the tests without changing the functionality.

tests/test_multiplanet.py
```diff
@@ -237,7 +236,7 @@ def test_fit_selfconsist():
     samp.curr_pos[0, 4] = mass_b
     samp.curr_pos[0, 5] = mass_c
 
-    samp.run_sampler(n_walkers * 50, burn_steps=50)
```
sblunt, line 240: This one started failing for me after I removed the `emcee` code in favor of `ptemcee` with `n_temps=1`. I think it needs either more steps or a wider tolerance for the mass of c.

PR #436, opened 2026-09-01
Fix slowdown when using periodic_save_freq in MCMC

sblunt:
periodic_save_freq previously re-derived the entire chain history (including an unvectorized per-sample prior recomputation) and rewrote the whole HDF5 output file on every periodic save, making cost grow with total run length instead of just the new chunk. _update_chains_from_sampler now only processes steps since the last save, and Results.save_results appends to resizable HDF5 datasets instead of rewriting the file each time.

(experimenting with claude)

### DIFF
--- a/orbitize/results.py
+++ b/orbitize/results.py
@@ -54,6 +54,13 @@ class Results(object):
         self._saved_filename = None
         self._n_rows_saved = 0
 
+        # backing buffers for add_samples(): allow post/lnlike to grow with
+        # amortized O(1) cost per row instead of reallocating+copying the
+        # full array on every call (see add_samples() docstring)
+        self._post_buf = None
+        self._lnlike_buf = None
+        self._n_used = 0
+
         if self.system is not None:
             self.tau_ref_epoch = self.system.tau_ref_epoch
             self.labels = self.system.labels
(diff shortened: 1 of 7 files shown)


sblunt:
Ok, Claude bungled this and I ended up tracking down the issue independently and writing a fix. Here are the things I changed (I made some other changes along the way):

- I removed logic for calling emcee and ptemcee separately. ptemcee operates exactly like emcee when n_temps = 1, so this change simplifies the code considerably.
- I changed the periodic save filenames so a new file is saved at each checkpoint. This should be slightly safer than the previous method-- if the code is interrupted during a save, the old checkpoints won't be corrupted.
- I removed the sampler.chain attribute. Currently, this attribute is just a copy of the ptemcee sampler chain attribute. As the mcmc runs and the chain gets bigger and bigger, having two copies of the chain in memory slows the run down considerably.

sblunt:
For completeness, the issue with the claude solution was that it focused on the wrong thing. The tallest tentpole here was the growth in memory, easily fixed by removing an unused variable, rather than the time it took to actually do the saving. I opted to revert the claude solution, even though it will made the code incrementally faster, because it makes the code a lot more complicated.

sblunt:
Ignore the failing multiplanet tests-- Eshel's got a PR in now to fix the seed for all the tests.

PR #437, opened 2026-09-15
Fix compute_model crash due to brightness when use_rebound=True

eshelDror:
Usually when using rebound it is set through `System.use_rebound` when initializing rather than the `use_rebound` keyword in `compute_model`. This error would only occur when the `use_rebound` keyword is used manually.


orbitize/system.py
```diff
@@ -638,9 +638,9 @@ def compute_model(self, params_arr, use_rebound=False):
         to_convert = np.copy(params_arr)
         standard_params_arr = self.basis.to_standard_basis(to_convert)
 
-        if use_rebound:
-            raoff, decoff, vz = self.compute_all_orbits(
-                standard_params_arr, comp_rebound=True
+        if use_rebound or self.use_rebound:
```
sblunt, line 641: Good catch. Let's just set `comp_rebound=True` so if the kwarg input is set to False but `self.use_rebound` is True, it still uses rebound.

sblunt: Retracting this comment based on irl conversation. This is a bit of a tangle, but we're doing this to avoid changing the behavior of the use_rebound keyword.

